PR #2, opened 2016-04-10
better integration of the "simple http server" AND Adobe-color profiles are now automatically downloaded

lrosenth:
I modified a few things:
1) better integration of the "simple http server" into sipi. Within a config file, I can add a route and docroot for serving normal http files (.html, .lua and .elua). I added a simple example which allows to upload and convert image files. The corresponding files are "upload.html" and "upload.elua" in the directory "server"

2) The Adobe-color profiles are now automatically downloaded during the build process (by cmake). I updated the Readme.md to reflect these changes and inform the user that he has to agree to the adobe licence conditions.


BiancoA:
Hi Lukas,

I gave a look at the code and fixed smaller bug that I will explain in another comment, I think we can merge this branch but I have a few comments to add:

- We should use names for the branches and pull requests that reflect the features that we are planning to add in that specific branch. In this way it will be easier to read the history in future;

- We should make a branch for each feature we want to add. For example this branch could be split in two smaller branches: httpServerIntegration and ColorProfileDownload . In this way the process of review can be more modular. In this case I fixed a small problem with the color profile download, but I haven't had a look at the httpServerIntegration yet. If it were two different branches I could have merged the ColorProfileDownload already;

- Branches after being downloaded are deleted, and till the pull request is merged no other mayor modifications should be made.

What do you think about it?

CMakeLists.txt
```diff
@@ -465,16 +465,19 @@ CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/SipiConfig.h.in ${CMAKE_CURRENT_SOURC
 # Here we get the ICC profiles from the Adobe website and prepare them for including
 # We just use the AdobeRGB1998.cc and the USWebCoatedSWOP.icc profiles for now...
 #
+unset(ICC CACHE)
```
BiancoA: this unset is needed otherwise the variable ICC is saved in CMakeCache.txt and not updated.

CMakeLists.txt
```diff
@@ -465,16 +465,19 @@ CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/SipiConfig.h.in ${CMAKE_CURRENT_SOURC
...
 find_file(
     ICC
-    ${CMAKE_CURRENT_SOURCE_DIR}/icc.zip
+    icc.zip
+    ${CMAKE_CURRENT_SOURCE_DIR}
     NO_DEFAULT_PATH
 )
-if(ICC-NOTFOUND)
```
BiancoA: this is explained in this [link](http://stackoverflow.com/questions/29657195/how-to-test-if-cmake-found-a-library-with-find-library)

CMakeLists.txt
```diff
@@ -486,14 +489,14 @@ add_custom_command(
         COMMENT "Generating ICC profile includes"
         WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
         DEPENDS icc.zip
-        COMMAND unzip ${CMAKE_CURRENT_SOURCE_DIR}/icc.zip -d ${CMAKE_CURRENT_SOURCE_DIR}
-        COMMAND cp '${CMAKE_CURRENT_SOURCE_DIR}/Adobe ICC Profiles \(end-user\)/RGB Profiles/AdobeRGB1998.icc' ${CMAKE_CURRENT_SOURCE_DIR}/
+        COMMAND unzip -f ${CMAKE_CURRENT_SOURCE_DIR}/icc.zip -d ${CMAKE_CURRENT_SOURCE_DIR}
```
BiancoA: to avoid being asked: "do you want to substitute existing file blabla" press [y] [n] ...

CMakeLists.txt
```diff
@@ -486,14 +489,14 @@ add_custom_command(
...
+	COMMAND cp '${CMAKE_CURRENT_SOURCE_DIR}/Adobe ICC Profiles \(end-user\)/RGB Profiles/AdobeRGB1998.icc' ${CMAKE_CURRENT_SOURCE_DIR}/
         COMMAND xxd -i AdobeRGB1998.icc AdobeRGB1998_icc.h
         COMMAND rm AdobeRGB1998.icc
-        COMMAND cp '${CMAKE_CURRENT_SOURCE_DIR}/Adobe ICC Profiles \(end-user\)/CMYK Profiles/USWebCoatedSWOP.icc' ${CMAKE_CURRENT_SOURCE_DIR}/
+	COMMAND cp '${CMAKE_CURRENT_SOURCE_DIR}/Adobe ICC Profiles \(end-user\)/CMYK Profiles/USWebCoatedSWOP.icc' ${CMAKE_CURRENT_SOURCE_DIR}/
         COMMAND xxd -i USWebCoatedSWOP.icc USWebCoatedSWOP_icc.h
         COMMAND rm USWebCoatedSWOP.icc
-        COMMAND rm -rf '${CMAKE_CURRENT_SOURCE_DIR}/Adobe ICC Profiles \(end-user\)'
+        COMMAND rm -rf '${CMAKE_CURRENT_SOURCE_DIR}/Adobe ICC Profiles *end-user*'
```
BiancoA: This is still not working... I haven't figured out why

PR #10, opened 2016-04-21
Add Config Files to test Sipi with Knora

tobiasschweizer:
Now, you can start Sipi with the config file `sipi.knora-test-config.lua`

It will always serve the same image and it won't do the roundtrip to Knora. Once we hace figured out how to automatically install the required Lua modules, I could add that too.


benjamingeer:
OK to merge!

PR #14, opened 2016-05-12
Fix brokenpipe

lrosenth:
the broken pipe bug should be fixed with this version. It required adding a own error handling to methods writing the JPEG file with libjeg. Otherwise libjpeg calls exit. In additions all exceotions in destructors have to the catched within the destructors. A exception raised in a destructor and propagted can never be catched!


BiancoA:
Thank you Lukas. 
I tried on Ubuntu 14.04 and it compiles and it works. I think you can merge it without any problem.

PR #23, opened 2016-05-27
Google test framework

BiancoA:
Google test framework is not introduced in this request, because I think that it is important to refactor the code first.

Config files and other relative path should be modified ( for example in the .lua files @tobiasschweizer )

@subotic  the tiff library is still giving strange problems: on my ubuntu it is now installed in local/lib/ while in travis it is under local/lib/x86_64-linux-gnu/ as it was on my previous linux installation. Strange.


BiancoA:
re-open, will delete after merge

PR #27, opened 2016-06-03
Tif16

lrosenth:
I fixed the bug with generating JPEG's from 16 bit images. Both in the command line version as in the HTTP-Server variant, requesting a JPEG from a 16 Bit image should create a valid JPEG by converting the image from 16 bits to 8 bits.

cc: @subotic @tobiasschweizer


BiancoA:
Hi Lukas,
thanks for the fix! Just one small thing: could you add to line 946 in file src/SipiImage.cpp a small comment describing what the code is doing? something like `bitwise right shift (equivalent of dividing by 2^8) and then masking the result to get the last 8 bit of the shifted int16 (to eliminate the zeros left)`

Have a good day!

lrosenth:
Hi Andrea,

is done… Shall I merge?

Lukas
--  

Prof. Dr. Lukas Rosenthaler
Head of "Digital Humanities Lab",
University of Basel
Bernoullistrasse 32
CH-4056 Basel, Switzerland

Tel: + 41 61 267 0490
Fax: +41 61 267 3855
http://www.dhlab.unibas.ch

BiancoA:
Perfect, Thanks!
Yes, I think you can merge.

PR #39, opened 2016-07-11
Docs/setup for centos

tobiasschweizer:
add documentation for setup on CentOS


tobiasschweizer:
@BiancoA what about this pull request?

BiancoA:
Ok! you can merge it. Presumably in a short time we should not need java anymore...

tobiasschweizer:
Presumably in a short time I will get very rich because I win a lot of money in a casino :-)

PR #56, opened 2016-09-13
cookie handling in test preflight request

tobiasschweizer:
I added the same cookie handling we use for the normal configuration for the use of Sipi with Knora.


BiancoA:
Thanks, you can merge it!

PR #61, opened 2016-09-28
Sqlite

lrosenth:
I integrated sqlite into SIPI. Now it's possible directly to query a sqlite database from SIPI (using Lua). Thus it would be possible to create a standalone image server with metadata using SIPI. In addition it should compile on most up-to-date Lines. Sorry with old versions (e.e. CentOS 6, old Ubuntu etc.

@BiancoA @schwemmer @tobiasschweizer


BiancoA:
Dear Lukas,

I checked the code and it seems ok to me. Since merging requires resolving some conflicts, could you run the tests? In case you already did it, I think you can merge it.

Andrea

PR #63, opened 2016-10-19
Keepalive

lrosenth:
The keep-alive header should now be processed correctly....


BiancoA:
Dear Lukas,

Thanks, I will have a look at the code tomorrow or Monday. Have a nice day!

BiancoA:
Dear Lukas, 

I checked the code, it can be merged.
The only comment: it seems to me that now the default option for error output is DEBUG, but maybe it should not be the default.
What do you think?

PR #65, opened 2016-10-21
fix (lua_table_to_json): treat lua numbers as integer or double

tobiasschweizer:
- check if a lua number has a fraction


lrosenth:
Yes, merge!! That's OK

PR #70, opened 2016-10-24
Tests/wrong req

BiancoA:
Added tests and source images for jpg2000. LeavesC is a lossy compressed image with ratio 1.0 bps and LeavesR is a lossless compressed image with 4 layers with ratio -,1.0, 0.5., 0.25 bps


lrosenth:
Please merge and pull! Looks good!!

PR #82, opened 2016-12-05
Non image file support

BiancoA:
This pull request:

- adds support for uploading non image files to the Sipi file server from a Lua script (see also new method `LuaServer.cpp:lua_fs_copyfile`)
- adds support for checking for an uploaded file's media type (determined from the mimetype) in Lua
- fix various bugs in the Knora Lua routes (new Lua error handling)
- add better error handling in `LuaServer.cpp:lua_copytmpfile` when accessing the vector representing the file uploads
- add `docroot` to the Lua config variable
- if a Lua runtime error occurs, print also the name of the function to the log
- fix error handling when calling the Lua pre-flight function, and refactor duplicated code
- refactor some code in `LuaServer.cpp` so it doesn't use pointers or arrays

We will move all the Knora related Lua config files and scripts to the Knora repository. However, if changes are made to `LuaServer.cpp`, it has to be checked if these files have to be adapted as well (unless the changes made are backward compatible).


lrosenth:
I'll merge it now!

PR #95, opened 2017-01-11
Img conversion bug

lrosenth:
I added support for all three IIIF quality classes (color, gray and bitonal) with the appropriate conversions. In addition bitonal TIFF's can be read, converted to JPEG2000, PNG (doesn't make sense but it works) and written (using FAX CCIT4 compression) as TIFF.

For the conversion from color and gray image, I implemented a Floyd-Steinberg dithering which seems to be a good compromise between quality and speed.

@BiancoA @benjamingeer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -304,9 +304,6 @@ add_executable(
         src/formats/SipiIOJ2k.cpp include/formats/SipiIOJ2k.h
         src/formats/SipiIOJpeg.cpp include/formats/SipiIOJpeg.h
         src/formats/SipiIOPng.cpp include/formats/SipiIOPng.h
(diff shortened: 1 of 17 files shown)


BiancoA:
Dear Lukas, 

Thanks for the request. very few comments:

- to compile zlib (ext/zlib/CMakeLists.txt) needs to be updated to zlib-1.2.11.tar.gz
- Before merging it would be better to handle the transition to the new kakadu also on Travis @subotic 
- all the current (beta) users of Sipi need to have a copy of the new kakadu library, otherwise it won't compile on their machines (@tobiasschweizer @loicjaouen @benjamingeer are those that I know)
- Could you provide a test image for the bitonal?
 
Andrea

Issue #85, opened 2016-12-13
If a config file is in the wrong format, Sipi crashes


benjamingeer:
```
$ local/bin/sipi -config config/sipi.init-knora.lua

Sipi Version 1.0 Beta (build 2016-12-12 13:57)
Sipi Source Git-Version: 5c89569
Segmentation fault: 11
```

BiancoA:
this is also kind of related: https://github.com/dhlab-basel/Sipi/issues/22

Issue #22, opened 2016-05-27
Failure in handling wrong command line options


BiancoA:
Passing a wrong option causes a segmentation fault:

```
andrea@Ubuntu-MacPro:~/Src/sipi$ build/sipi -c sipi.config.lua 
Segmentation fault (core dumped)
andrea@Ubuntu-MacPro:~/Src/sipi$ build/sipi -c 3
Segmentation fault (core dumped)
andrea@Ubuntu-MacPro:~/Src/sipi$ build/sipi -c w
Segmentation fault (core dumped)
andrea@Ubuntu-MacPro:~/Src/sipi$ build/sipi -a w
Segmentation fault (core dumped)
```

benjamingeer:
Could be fixed by using [GNU getopt](http://www.gnu.org/software/libc/manual/html_node/Getopt.html).

benjamingeer:
A problem with GNU getopt could be that, on Mac OS X, there is already BSD's getopt, which doesn't support long options. GNU getopt can be installed with Homebrew (`brew install gnu-getopt`), but it's not linked into `/usr/local`, to keep it from interfering with the system getopt, so I'm not sure how to get CMake to use it.

We could try this getopt implementation, which seems similar to GNU getopt but is provided as a single header file, which we could add to the  Sipi git repository:

http://optionparser.sourceforge.net/

BiancoA:
working on it here: https://github.com/dhlab-basel/Sipi/tree/fix/comLineOptions

PR #98, opened 2017-01-16
Fix/com line options

BiancoA:
new command line library, POSIX compliant

Does:
- start SIPI server from a config file
- start SIPI server using command line arguments
- diff between two images
- check that the file exists

Missing:
- check that the lua file is a correct lua file

Solves:
- https://github.com/dhlab-basel/Sipi/issues/22

Maybe could solve:
- https://github.com/dhlab-basel/Sipi/issues/85


lrosenth:
I have one problem: the output filename "-" has been used with a special meaning: write the data to stdout. However the new command line parser is not able to deal with a "-" as a special filename...
I will have to modify the internal file handling to use another special filename for stdout (maybe "stdout:"...
I'll make this modification in this branch...

lrosenth:
OK, now You can merge...

PR #104, opened 2017-01-20
Fix/libmagic fix

lrosenth:
I "outsourced" libmagic since it posed problems with hard-wired location of the "magic" file. SIPI now expects that libmagic is installed by yum (linux) or brew (os x)


lrosenth:
@subotic Could add `sudo yum install file-devel` to the travis-thingy? We need libmagic installed on the system (I had to take out lib magic from SIPI because there was an absolute path to the SIPI level directory...

CMakeLists.txt
```diff
@@ -329,10 +329,13 @@ add_dependencies(sipi icc_profiles)
 
 
 # MESSAGE(STATUS "LIBS: " ${LIBS})
-target_link_libraries(sipi ${LIBS} lcms2 exiv2 expat jpeg tiff jbigkit png kdu_aux kdu xz magic lua jansson sqlite3 dl pthread curl ${CMAKE_DL_LIBS} z m )
+target_link_libraries(sipi ${LIBS} lcms2 exiv2 expat jpeg tiff jbigkit png kdu_aux kdu xz magic lua jansson sqlite3 dl pthread curl ${CMAKE_DL_LIBS} z m)
 if(CMAKE_SYSTEM_NAME STREQUAL DARWIN)
 	target_link_libraries(sipi iconv)
+else()
+    target_link_libraries(sipi rtmp rt)
```
benjamingeer: The library `librtmp` doesn't seem to be available as a binary package on CentOS. This is preventing Sipi from compiling on that platform. I've tried to get CMake to download `librtmp` and compile it, but without success, because `librtmp` seems to require OpenSSL (which Sipi doesn't), and in any case `librtmp` doesn't find the OpenSSL headers even if they're installed. Not sure what to do about this. Why do we need this library?

lrosenth:
I really do not know which module needs librtmp... I would like to get rid of it totally

benjamingeer:
OK, I removed all references to `librtmp `, and now Sipi builds and works on Mac OS X, Ubuntu, and CentOS. So I think we can merge this pull request.

Issue #107, opened 2017-01-24
Dockerfile


loicjaouen:
where should we put `Dockerfile`s ?
(I have one for `centos` that builds sipi, and one that runs sipi for knora)

subotic:
I depends ;-)

If you create dockerfiles that are customised for your deployment, then you should have a private git repository to store them.

The public general use dockerfile for Sipi is in the root of the sipi repository. Knora doesn't have one yet. The ones for GraphDB and Fuseki are in their respective folders.

PR #108, opened 2017-01-25
Dockerfile for building Sipi on centos7

loicjaouen:
as discussed on:
https://github.com/dhlab-basel/Sipi/issues/107


subotic:
Why do we need this? There is already a ``Dockerfile`` inside the repository. If you would like to create your own custom images, then those Dockerfiles should be stored somewhere else.

Also, @lrosenth is working on a new version, which will be used to produce the official Sipi docker image hosted on the docker hub.

loicjaouen:
I need it because it self contained: it builds Sipi (the one provided now is packaging a local Sipi build).
Alternatively, it is centos based (current one is based on your ubuntu sipi-base image).

I like it because on prod, I can grab that file only and get rolling.

And I think it might be useful to others too :)
So back to #107 , if not there, do you want it somewhere?

subotic:
In PR #104 I have changed the base image that is used. Now images from the ``dhlabbasel`` organisation are used.

The ``dhlabbasel:sipi-base`` image is hosted on docker hub, and is automatically rebuilt on every commit to [docker-sipi-base](https://github.com/dhlab-basel/docker-sipi-base). We also have [docker-sipi](https://github.com/dhlab-basel/docker-sipi)  which you could use to store your Dockerfiles. You could put it there in a subdirectory.

subotic:
I have split up the Dockerfile into two (``dhlabbasel/sipi-base`` and ``dhlabbasel/sipi``), because the building of ``dhlabbasel/sipi-base`` takes a long time. We can easily extend ``docker-sipi-base`` and add a directory with a Dockerfile that creates a CentOS based image and then tag it accordingly. Then you could use in your ``sipi`` Dockerfile something like: ``FROM dhlabbasel/sipi-base:centos``.

Unfortunately, we cannot use docker hub to create automatic builds of ``sipi`` because of the ``kakadu`` library. Here the idea is to do this on Travis, and then push a clean image (with the kakadu source removed) to ``dhlabbasel/sipi``.

loicjaouen:
if I get it right, the answers are:
- dockerfiles should be put in https://github.com/dhlab-basel/docker-sipi-base
- this dockerfile doesn't fit the official build plan, it is then custom and should indeed be stored privately

Here it is:
https://github.com/LaDHUL/dockerimages

PR #110, opened 2017-01-25
Use smart pointers for some objects

benjamingeer:
This fixes some memory leaks identified by valgrind, and introduces some smart pointers to help prevent them in the future.

* Before shutting down, free memory allocated on startup.
* Fix some leaks that could occur when exceptions are thrown.
* Use `std::shared_ptr` for:
  * some member variables of `SipiImage`
  * `SipiSize`
  * `SipiRegion`
  * `SipiCache`
* Use `std::unique_ptr` for `SockStream` in `Server`.
* Fix a mutex deadlock that could occur on shutdown in `Server`.
* When copying a `SipiImage`, copy its members.
* Get rid of the global `Server` pointer, which was always passed from `main()` anyway.
* Replace `mktemp()` with the safer `mkstemp()`.
* Replace `NULL` with the safer keyword `nullptr` from C++11.
* Use `auto` to eliminate some compiler warnings about type conversions.
* Replace some C-style casts with `static_cast`.
* Rename some variables that were shadowing other variables.
* Rename some variables that were called `tmp`.

Using the smart pointers doesn't seem to have any effect on performance.

I put some questions in TODOs in `Connection.cpp`, lines 478-80.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,7 +323,7 @@ add_executable(
         shttps/GetMimetype.cpp shttps/GetMimetype.h
         shttps/Server.cpp shttps/Server.h
         shttps/jwt.c shttps/jwt.h
-)
+        shttps/makeunique.h)
 add_dependencies(sipi icc_profiles)
 
 
--- a/include/SipiCache.h
+++ b/include/SipiCache.h
@@ -178,11 +178,11 @@ namespace Sipi {
         std::string check(const std::string &origpath_p, const std::string &canonical_p);
 
         /*!
-         * generate a new name for a cache file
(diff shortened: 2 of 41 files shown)


lrosenth:
@benjamingeer **Hi Ben, please merge! Everything OK**.
TODO in Connections: I want the strings to be '\0'-terminated, therefore this complex "construct". The Chunkreader allocates the memory (using realloc!). Therefore we do it like we do it ;-) and free tmp afterwards [Unfortunately C++ has now "realloc" which is more efficient that malloc/free only. If there is space it is able to add memory without copying the existing parts...

PR #111, opened 2017-01-31
Use std::unique_ptr for some char arrays

benjamingeer:
This aims to fix/prevent memory leaks by using `std::unique_ptr` to manage some short-lived `char` arrays, making it unnecessary to call `delete` explicitly.

It also removes some code that is no longer needed.

I've added some comments below to indicate places where it seemed that memory was allocated but not freed. @lrosenth, could you please check if I've understood correctly?


shttps/Hash.cpp
```diff
@@ -90,19 +91,19 @@ namespace shttps {
     //==========================================================================
 
     bool Hash::hash_of_file(const string &path, size_t buflen) {
-        char *buf = new char[buflen];
```
benjamingeer: As far as I could tell, this memory was not freed.

src/metadata/SipiIcc.cpp
```diff
@@ -41,12 +41,12 @@ namespace Sipi {
             throw SipiError(__file__, __LINE__, "cmsOpenProfileFromMem failed");
         }
         unsigned int len = cmsGetProfileInfoASCII(icc_profile, cmsInfoDescription, cmsNoLanguage, cmsNoCountry, nullptr, 0);
-        char *buf = new char[len];
```
benjamingeer: As far as I could tell, this memory was not freed.

src/metadata/SipiIcc.cpp
```diff
@@ -172,12 +167,11 @@ namespace Sipi {
     SipiIcc& SipiIcc::operator=(const SipiIcc &rhs) {
         if (this != &rhs) {
             if (rhs.icc_profile != nullptr) {
-                char *buf = nullptr;
                 unsigned int len = 0;
                 cmsSaveProfileToMem(rhs.icc_profile, nullptr, &len);
-                buf = new char[len];
```
benjamingeer: As far as I could tell, this memory was not freed.

lrosenth:
@benjamingeer Hi Ben, I successfully compiled it and tested it (during the talks...). Go ahead and merge!! Looks very nice!

PR #114, opened 2017-02-02
Add support for text files

tobiasschweizer:
This pull requests adds support for uploading text files. 
- add a function that handles the mimetype and charset information provided by the user (`split_mimetype_and_charset`)

It also fixes some bugs:
- if the preflight_request return restricted permission, the the filepath was not set, leading to a "FILE NOT FOUND" error -> https://github.com/dhlab-basel/Sipi/pull/114/commits/dfd2a9fbf1363a86debd30424c3aafe1ae785eee
- the charset contained a leading space -> https://github.com/dhlab-basel/Sipi/pull/114/commits/9a1fc466dbcd1c5aefe0993c0828bc0da56049bb

@lrosenth Could you have a look at my changes?


tobiasschweizer:
@benjamingeer Do you think we should also test our Lua scripts?

benjamingeer:
Yes, we should test the Lua scripts.

If you type `make check`, several tests are run that use `sipi.fake-knora-test-config.lua`, using `nginx` to simulate Knora. Currently `nginx` always returns the same permission, so the result is always `allow`. We should add more tests to cover `restrict` and `deny`.

tobiasschweizer:
> If you type make check, several tests are run that use sipi.fake-knora-test-config.lua, using nginx to simulate Knora. Currently nginx always returns the same permission, so the result is always allow. We should add more tests to cover restrict and deny.

Yes :-)

How would you propose to test the Lua scripts (i.e the functions, not the routes)? Testing them in Lua would be easy:

```
require "Luafile"

-- make function calls with params and check results
```

Or would you like to call Lua from Python?

tobiasschweizer:
there is this <http://labix.org/lunatic-python> but probably it is a dead project

benjamingeer:
If you want to test that your `pre_flight` function works correctly, you can make `nginx` routes that return different permissions for different filenames. Then you can use the Python-based test framework to make HTTP requests for those files, and check that Sipi returns the expected HTTP status codes.

To test all the Sipi functions that can be called from Lua scripts, we could make test routes in Sipi. Each route would call a Sipi function and return some evidence that it worked (e.g. in an HTTP response header). These routes could then be tested using the Python-based framework.

benjamingeer:
To test `restrict`, you could check the image returned by Sipi to see if it has the expected watermark.

tobiasschweizer:
Ok, I get it.

The scenario if was referring to would be an additional thing: tests the Lua Lua functions that are called by the Lua routes.

For example, there are functions that do string processing to handle mime types provided by the client etc.

benjamingeer:
> For example, there are functions that do string processing to handle mime types provided by the client etc. So we could write a test for this very function with different parameters without having to to tests the same route several times, while just changing the mime type string.

Having put those tests in a Lua function, you could then make a Sipi route that calls that function. If the tests pass, it returns 200, otherwise 500.

tobiasschweizer:
Ok, perfect. Let's do it then.

scripts/get_mediatype.lua
```diff
@@ -47,31 +47,75 @@ function get_mediatype(mimetype)
...
+
+        -- original mimetype also contains the encoding, e.g. "application/xml; charset=UTF-8"
+        -- split the string using the position information start_char and end_char
+
+        -- get a substring from the beginning until the position of "; "
+        -- substract 1 from start_char because Lua includes the last character which we do not want
+        local mimetype = string.sub(mimetype_and_charset, 1, start_char -1) -- e.g. "application/xml"
+
+        -- get a substring from the end of "; " until the last character
+        -- add 1 to end_char because otherwise Lua includes the whitespace which we do not want
```
benjamingeer: The whitespace after the semicolon is optional: https://tools.ietf.org/html/rfc7231#section-3.1.1

benjamingeer:
I'm still getting 404 Not Found for `restrict`. You can test it like this:

```
nginx -p test/nginx -c nginx.conf
local/bin/sipi --config config/sipi.fake-knora-test-config.lua
```

Then try to access http://localhost:1024/knora/RestrictLeaves.jpg/full/full/0/default.jpg. Sipi says `Not Found`.

tobiasschweizer:
This is because there is no file "RestrictLeaves.jpg" in `./test/_test_data/images`.

Could you try <http://localhost:1024/knora/Leaves8.jp2/full/full/0/default.jpg>?

benjamingeer:
It works!

benjamingeer:
This compiles on CentOS but the regex for parsing MIME types doesn't work. Perhaps a problem with gcc.

benjamingeer:
It may help to try a newer gcc on CentOS, available in Devtoolset-4:

https://www.softwarecollections.org/en/scls/rhscl/devtoolset-4/

benjamingeer:
I just need to get this working on CentOS. I'll try this afternoon.

tobiasschweizer:
I will install Virtualbox and also run CentOS 7 as a VM.

tobiasschweizer:
I have forgotten how much I hate CentOS.

benjamingeer:
It takes 10 minutes to set up Ubuntu in VirtualBox, and two hours to set up CentOS. :(

tobiasschweizer:
How do I get `xxd`?

benjamingeer:
It's in `vim-common`, but how do you know you need it? I haven't needed to install it.

tobiasschweizer:
I think it is needed to build the color profiles

benjamingeer:
Did you actually get an error message about it?

tobiasschweizer:
yes, but I got so angry at CentOS that I deleted the whole VM and downloaded the everything iso

tobiasschweizer:
did you manage to set up Guest Additions for Virtualbox on Mac OS?

benjamingeer:
Yes. I used the everything ISO and followed the instructions here (except that I didn't bother with the `yum-priorities` plugin):

https://wiki.centos.org/HowTos/Virtualization/VirtualBox/CentOSguest

Then installed this to get gcc 5.3:

https://www.softwarecollections.org/en/scls/rhscl/devtoolset-4/

And did this to make bash find gcc 5.3 by default on login:

http://unix.stackexchange.com/a/195219

And then followed the current Sipi build instructions for CentOS. Now this branch compiles and runs correctly (including the regex stuff). If you have time, you could update the build docs accordingly, otherwise I'll do it tomorrow.

benjamingeer:
Forgot to mention: after you follow these instructions:

https://wiki.centos.org/HowTos/Virtualization/VirtualBox/CentOSguest

You have to do `yum update` and reboot.

tobiasschweizer:
@lrosenth @BiancoA Are you ok with the requirement of gcc >= 5.3?

PR #118, opened 2017-02-06
Show more readable command-line option syntax

benjamingeer:
This just changes the command-line option help to suggest these

```
--config foo
-c foo
```

instead of these, which I think aren't as readable:

```
---config=foo
-cfoo
```


BiancoA:
it works fine for me, i think you can merge!

PR #126, opened 2017-02-14
Improve error checking in SipiSize and SipiRotation

### LINKED ISSUE #97: IIIF validation sometimes fails
```
           [18] test size_error_random FAIL
             exception: Remote end closed connection without response
```

This is the code for that test in the IIIF validator:

https://github.com/IIIF/image-validator/blob/master/iiif_validator/tests/size_error_random.py

### LINKED ISSUE #122: When given an incorrect IIIF rotation parameter, Sipi doesn't return an error
One of the IIIF validation tests puts garbage in the rotation parameter, and expects to get a 400 Bad Request error, but Sipi accepts the request and returns a rotated image.

To reproduce:

```
nginx -p test/nginx -c nginx.conf
local/bin/sipi --config config/sipi.fake-knora-test-config.lua
```

Then load this URL in a browser:

http://localhost:1024/knora/67352ccc-d1b0-11e1-89ae-279075081939.jp2/full/full/5Ez%3E/default.jpg

benjamingeer:
This fixes two IIIF validation errors, so the IIIF validation tests always pass:

* Fixes #97.
* Fixes #122.

Summary of changes:

* Change the constructors of `SipiSize` and `SipiRotation` to make Sipi return a 400 Bad Request error when the IIIF size or rotation parameters are invalid.
* Collect parsing functions for integers, floats, and MIME types in `shttps::Parsing`.
* Construct a `SipiSize` on the stack instead of on the heap in `SipiLua.cpp`.
* Replace some `int` with `size_t`.
* Replace some C-style casts with `static_cast` (especially in `SipiSize.cpp`).
* Reformat some code for readability.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,7 +318,7 @@ add_executable(
         shttps/Connection.cpp shttps/Connection.h
         shttps/LuaServer.cpp shttps/LuaServer.h
         shttps/LuaSqlite.cpp shttps/LuaSqlite.h
-        shttps/GetMimetype.cpp shttps/GetMimetype.h
+        shttps/Parsing.cpp shttps/Parsing.h
         shttps/Server.cpp shttps/Server.h
         shttps/jwt.c shttps/jwt.h
         shttps/makeunique.h)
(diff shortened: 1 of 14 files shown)


lrosenth:
Wonderful!!

PR #130, opened 2017-02-17
Fix/chunked post

### LINKED ISSUE #121: Sipi can't receive a chunked POST request, and crashes if Content-Length is too large
If I send a file upload as chunked POST request to Sipi (e.g. to `/make_thumbnail`), Sipi doesn't process the upload. This is because it expects a `Content-Length` header, which is not part of a chunked request. A chunked request contains the header `Transfer-Encoding: chunked` instead. The HTTP 1.1 spec [says](https://tools.ietf.org/html/rfc2616#section-4.4):

> If a Content-Length header field (section 14.13) is present, its decimal value in OCTETs represents both the entity-length and the transfer-length. The Content-Length header field MUST NOT be sent if these two lengths are different (i.e., if a Transfer-Encoding header field is present). If a message is received with both a Transfer-Encoding header field and a Content-Length header field, the latter MUST be ignored.

To reproduce, start Sipi like this:

```
local/bin/sipi --config config/sipi.knora-config.lua
```

Then make a request like this:

```
curl -v -i -X POST -H "Content-Type: multipart/form-data" -H "Transfer-Encoding: chunked" --form "file=@test/_test_data/images/knora/Leaves.jpg" http://localhost:1024/make_thumbnail
```

The `make_thumbnail.lua` script returns an error saying `no image uploaded`.

The problem starts on line 474 of `Connection.cpp`:

```
                if (content_length > 0) {
```

Sipi checks whether the request is chunked only after checking the content length.

Also, the code there passes the value of the `Content-Length` header to `malloc`, without checking whether that's a reasonable value. This means that you can make Sipi crash by making a POST request with a very large content length (even without uploading a file):

```
curl -v -i -X POST -H "Content-Type: multipart/form-data" -H "Content-Length: 100000000000000" --form "file=@test/_test_data/images/knora/Leaves.jpg" http://localhost:1024/make_thumbnail
```

@lrosenth Would you consider replacing Sipi's HTTP server code with https://github.com/facebook/proxygen ?

lrosenth:
@benjamingeer @BiancoA 
Dealing with chunked uploads should work now. I also added a max_post_size parameter to the Lua config file in order to prevent unlimited uploads. Setting

max_post_size = '100M'

restricts uploads to a maximum of 100MBs. is max_post_soize ='0', then the parameter is disabled and unlimited uploads are allowed (not recommended ;-) but may be used for test purposes

Fixes #121.


shttps/ChunkReader.cpp
```diff
@@ -49,10 +49,16 @@ using namespace std;
...
+    };
+    //=========================================================================
+
+    size_t ChunkReader::read_chunk(istream &ins, char **buf, size_t offs)
     {
         string line;
-        (void) safeGetline(ins, line); // read empty line
+        (void) safeGetline(ins, line); // read chunk size
         if (ins.fail() || ins.eof()) {
             throw -1;
```
benjamingeer: Could this throw `Error` instead of -1?

shttps/ChunkReader.cpp
```diff
@@ -79,7 +94,7 @@ namespace shttps {
...
```
benjamingeer: Same

shttps/ChunkReader.cpp
```diff
@@ -79,7 +94,7 @@ namespace shttps {
...
         }
         (*buf)[offs + n] = '\0';
-        (void) safeGetline(ins, line); // read empty line
+        (void) safeGetline(ins, line); // read "\r\n" at end of chunk...
         if (ins.fail() || ins.eof()) {
             throw -1;
```
benjamingeer: Here too

shttps/ChunkReader.cpp
```diff
@@ -108,13 +125,11 @@ namespace shttps {
...
-        std::istream::sentry se(*ins, true);
-        std::streambuf* sb = ins->rdbuf();
         size_t n = 0;
         for(;;) {
             if (chunk_pos >= chunk_size) {
                 string line;
-                (void) safeGetline(*ins, line);
+                (void) safeGetline(*ins, line); // read chunk size
                 if (ins->fail() || ins->eof()) {
                     throw -1;
```
benjamingeer: Same

shttps/ChunkReader.cpp
```diff
@@ -49,10 +49,16 @@ using namespace std;
...
+        chunk_size = 0;
+        chunk_pos = 0;
+    };
+    //=========================================================================
+
+    size_t ChunkReader::read_chunk(istream &ins, char **buf, size_t offs)
     {
         string line;
-        (void) safeGetline(ins, line); // read empty line
+        (void) safeGetline(ins, line); // read chunk size
```
benjamingeer: Is the comment here correct? What happens if the line is 1 terabyte long?

shttps/ChunkReader.cpp
benjamingeer: (There are several more of these in this file.)

shttps/Connection.cpp
```diff
@@ -471,21 +470,24 @@ namespace shttps {
...
+                    if (_chunked_transfer_in) {
+                        char *chunk_buf;
+                        ChunkReader ckrd(ins, _server->max_post_size());
+                        content_length = ckrd.readAll(&chunk_buf);
+                        if ((bodybuf = (char *) malloc((content_length + 1) * sizeof(char))) == nullptr) {
+                            throw Error(__file__, __LINE__, "malloc failed!", errno);
                         }
-                        else {
+                        memcpy(bodybuf, chunk_buf, content_length);
+                        free(chunk_buf);
```
benjamingeer: It looks like the `chunk_buf` won't be freed if the exception is thrown in this block. How about changing `ChunkReader` to take a `std::string` as a buffer instead?

shttps/ChunkReader.cpp
lrosenth: No – this is finally not an error. It just means that the client has cancelled the connection. But I could make a typedef to have a better name for this case...

shttps/ChunkReader.cpp
lrosenth: I'll have a look at this....

shttps/Connection.cpp
lrosenth: Right with exceptions. Since I'm using malloc and realloc which is more efficient, since it just increases the heap block if possible and copies the data only to the end of the heap if the block is too small. Thus it helps against fragmentation of the heap and is often much faster. String would always internally copy to a new heap location.

shttps/ChunkReader.cpp
lrosenth: a) Yes, it reads the chunk size...

b) I added a size check to safeGetline. Now an error is thrown if a line to too long (header lines are restricted to 64k, all others to max_post_size.

shttps/ChunkReader.cpp
lrosenth: Ok, replaces the "throw -1" with a typedef of an enum... Much better indeed ti understand what happens ;-)

shttps/Connection.cpp
benjamingeer: In that case, how about using a `shared_ptr` or `unique_ptr` so you don't have to remember to call `free`?

benjamingeer:
I see you replaced `throw -1` with something much clearer in `Connection.cpp`, could you do the same thing in `ChunkReader.cpp`?

shttps/Connection.cpp
benjamingeer: Another idea: could the pointer to the buffer be a member variable in the `ChunkReader` class? Then `ChunkReader` could free the memory in its destructor. Then there would be no need for the caller to free it.

lrosenth:
Ok, ChunkReader is now clean of "throw -1"'s – merge??

benjamingeer:
This still makes Sipi crash:

```
curl -v -i -X POST -H "Content-Type: multipart/form-data" -H "Content-Length: 100000000000000" --form "file=@test/_test_data/images/knora/Leaves.jpg" http://127.0.0.1:1024/make_thumbnail
```

The log says:

```
Feb 21 15:40:01  Sipi[19594] <Info>: Accepted connection from 127.0.0.1
libc++abi.dylib: terminating with uncaught exception of type std::out_of_range: stoi: out of range
Abort trap: 6
```

lrosenth:
OK, I see – that’s an error converting the string 100000000000000 to an integer… Overlooked this… Will fix it immediately ;-)

Lukas

On 21 Feb 2017, at 15:41, Benjamin Geer <notifications@github.com<mailto:notifications@github.com>> wrote:


This still makes Sipi crash:

curl -v -i -X POST -H "Content-Type: multipart/form-data" -H "Content-Length: 100000000000000" --form "file=@test/_test_data/images/knora/Leaves.jpg" http://127.0.0.1:1024/make_thumbnail


The log says:

Feb 21 15:40:01  Sipi[19594] <Info>: Accepted connection from 127.0.0.1
libc++abi.dylib: terminating with uncaught exception of type std::out_of_range: stoi: out of range
Abort trap: 6

benjamingeer:
I also need to try again to run the Knora-Sipi integration test that led to this bug report, and see if it works now.

lrosenth:
OK, fixed (and tested!)

benjamingeer:
I've been testing this with the Knora-Sipi integration tests that I've been writing, and I think something is still not quite right. My test tries to post a chunked request, using Akka HTTP, to Sipi's `/make_thumbnail` route. I put a lot of debugging output in `Connection.cpp`, and I've traced the problem to this code (line 725):

```
                                inbyte = _chunked_transfer_in ? ckrd.getc() : ins->get(); // get '\n';
                                if (ins->fail() || ins->eof()) {
                                    throw INPUT_READ_FAIL;
                                }
```

This code is executed after Sipi has read the whole uploaded file from the client, finished writing the temporary file, and read the last boundary. In my test, `inbyte` is EOF (-1), so there is no `\n`, but Sipi doesn't check for this. Therefore, Sipi never returns a response to the client. I'm not sure how to fix this.

lrosenth:
OK, let’s have a look at it… Are Youavailable  on Skype tomorrow?

Lukas
Am 22.02.2017 um 00:43 schrieb Benjamin Geer <notifications@github.com<mailto:notifications@github.com>>:


I've been testing this with the Knora-Sipi integration tests that I've been writing, and I think something is still not quite right. My test tries to post a chunked request, using Akka HTTP, to Sipi's /make_thumbnail route. I put a lot of debugging output in Connection.cpp, and I've traced the problem to this code (line 725):

                                inbyte = _chunked_transfer_in ? ckrd.getc() : ins->get(); // get '\n';
                                if (ins->fail() || ins->eof()) {
                                    throw INPUT_READ_FAIL;
                                }


This code is executed after Sipi has read the whole uploaded file from the client, finished writing the temporary file, and read the last boundary. In my test, inbyte is EOF (-1), so there is no \n, but Sipi doesn't check for this. Therefore, Sipi never returns a response to the client. I'm not sure how to fix this.

benjamingeer:
Great, I'll be available on Skype all day tomorrow.

benjamingeer:
Also I get this in Sipi's log:

```
Feb 22 00:46:25  Sipi[34901] <Warning>: Error shutting down socket at [/Users/benjamingeer/git/Sipi/shttps/Server.cpp: 661]: Socket is not connected
```

lrosenth:
Wonderful – I see a solution to it. We just have to check if there is a „dangling“ \n available or not. The error message is a result of throwing an exception if this last read fails.

According the specs the last boundary should be terminated by \r\n, but I’ll check if this is missing or if I read it before (which I suspect ;-)

Have a good night,
Lukas


Encoded data[edit<https://en.wikipedia.org/w/index.php?title=Chunked_transfer_encoding&action=edit&section=6>]
In the following example, three chunks of length 4, 5 and 14 are shown. The chunk size is transferred as a hexadecimal number followed by \r\n as a line separator, followed by a chunk of data of the given size.

4\r\n
Wiki\r\n
5\r\n
pedia\r\n
E\r\n
 in\r\n
\r\n
chunks.\r\n
0\r\n
\r\n


Note: the chunk size indicates the size of the chunk data and excludes the trailing CRLF ("\r\n").[2]<https://en.wikipedia.org/wiki/Chunked_transfer_encoding#cite_note-2> In this particular example, the CRLF following "in" is counted toward the chunk size of 0xE (14). The CRLF in its own line is also counted toward the chunk size. The period character at the end of "chunks" is the 14th character, so it is the last data character in that chunk. The CRLF following the period is the trailing CRLF, so it is not counted toward the chunk size of 0xE (14).

Am 22.02.2017 um 00:46 schrieb Benjamin Geer <notifications@github.com<mailto:notifications@github.com>>:


Great, I'll be available on Skype all day tomorrow.

lrosenth:
By the way,

Can You send me

1) how You start Sipi (which config file)?
2) the http request which produces the problem?

This would help me a lot…

Lukas

Am 22.02.2017 um 00:43 schrieb Benjamin Geer <notifications@github.com<mailto:notifications@github.com>>:


I've been testing this with the Knora-Sipi integration tests that I've been writing, and I think something is still not quite right. My test tries to post a chunked request, using Akka HTTP, to Sipi's /make_thumbnail route. I put a lot of debugging output in Connection.cpp, and I've traced the problem to this code (line 725):

                                inbyte = _chunked_transfer_in ? ckrd.getc() : ins->get(); // get '\n';
                                if (ins->fail() || ins->eof()) {
                                    throw INPUT_READ_FAIL;
                                }


This code is executed after Sipi has read the whole uploaded file from the client, finished writing the temporary file, and read the last boundary. In my test, inbyte is EOF (-1), so there is no \n, but Sipi doesn't check for this. Therefore, Sipi never returns a response to the client. I'm not sure how to fix this.

benjamingeer:
OK, I simplified the test to make it easier to reproduce.

1. In the Sipi source tree:
```
mkdir images
local/bin/sipi --config config/sipi.knora-config.lua
```
2. In the Knora git repository, check out the branch `wip/sipi-integration-tests`, then:
```
cd Knora/webapi
sbt
it:test-only *SipiV1ITSpec
```

The Knora test says:

```
[info] - Check if SIPI is running (2 seconds, 348 milliseconds)
[info] Knora and Sipi
[info] - should create a thumbnail *** FAILED *** (10 seconds, 34 milliseconds)
[info]   java.util.concurrent.TimeoutException: Futures timed out after [10 seconds]
```

Sipi says:

```
Feb 22 09:40:13  Sipi[36734] <Info>: Accepted connection from 127.0.0.1
A) -
A) lastboundary
#730 45
#732 -1
```

benjamingeer:
From [RFC 2616](https://tools.ietf.org/html/rfc2616#section-3.6.1), this is the ABNF for chunked encoding:

```
       Chunked-Body   = *chunk
                        last-chunk
                        trailer
                        CRLF

       chunk          = chunk-size [ chunk-extension ] CRLF
                        chunk-data CRLF
       chunk-size     = 1*HEX
       last-chunk     = 1*("0") [ chunk-extension ] CRLF

       chunk-extension= *( ";" chunk-ext-name [ "=" chunk-ext-val ] )
       chunk-ext-name = token
       chunk-ext-val  = token | quoted-string
       chunk-data     = chunk-size(OCTET)
       trailer        = *(entity-header CRLF)
```

Here's an example:

https://en.wikipedia.org/wiki/Chunked_transfer_encoding#Encoded_data

If I understand this correctly, a `Chunked-Body` has to end with:

```
\r\n    <- trailing CRLF of next-to-last chunk
0\r\n   <- last chunk and trailing CRLF of last chunk
\r\n    <- trailing CRLF of Chunked-Body
```

Is that different from what you're seeing?

benjamingeer:
OK, I think you're right that it has to do with the last multipart delimiter. According to [RFC 1341](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html), the structure of a multipart request is like this:

```
multipart-body := preamble 1*encapsulation 
               close-delimiter epilogue 

encapsulation := delimiter CRLF body-part 

delimiter := CRLF "--" boundary   ; taken from  Content-Type field. 
                                  ; when content-type is multipart 
                                  ; There must be no space 
                                  ; between "--" and boundary. 

close-delimiter := delimiter "--" ; Again, no  space  before "--" 

preamble :=  *text                ;  to  be  ignored  upon receipt. 

epilogue :=  *text                ;  to  be  ignored  upon receipt. 

body-part = <"message" as defined in RFC 822, 
         with all header fields optional, and with the 
         specified delimiter not occurring anywhere in 
         the message body, either on a line by itself 
         or as a substring anywhere.  Note that the 
         semantics of a part differ from the semantics 
         of a message, as described in the text.> 
```

Note that the `close-delimiter` is not supposed to have a trailing CRLF.

With `nc`, the end of the request looks like this (as viewed in Emacs):

multipart

`1e` (30 bytes) is the length of the chunk containing the `close-delimiter`, which ends with `--`. After that, there's a trailing CRLF for the chunk, then the empty chunk, then the trailing CRLF of the `Chunked-Body`.

So I think the request is correct, and Sipi should not expect a trailing CRLF after the `close-delimiter`.

benjamingeer:
There's still a problem: Sipi always thinks the submitted MIME type is `text/plain`, regardless of the MIME type that was submitted. Sipi's log output says:

```
LINE=Content-Type: image/jpeg
LINE=Content-Disposition: form-data; filename="Chlaus.jpg"; name="file"
File: Chlaus.jpg mimetype: text/plain
```

Here you can see that even though Sipi received `image/jpeg`, it decided that the MIME type was `text/plain`. This seems to be because of `Connection.cpp` line 596:

```
                        mimetype = "text/plain";
```

benjamingeer:
I removed that line and now everything seems to work. So I think you can remove the debugging output and merge this pull request.

PR #132, opened 2017-02-23
add build script for gh-pages

tobiasschweizer:
`rm -rf ../documentation` at the end of the script is needed because git keeps files under version control but not directories. So the directory is left over when switching back to `develop` which does not exist on this branch.

I tested the last steps (`git commit` and `git push` manually). When I wanted to rerun the script, it aborts because `git commit` without changes to the docs git returns -1 and the script stops (setting `set -e`).


tobiasschweizer:
Ok, I temporarily commented out `set -e` and rerun it. Seems to be working fine.

tobiasschweizer:
The option `set -e` was never commented out on this branch, I only did that locally in order to test the commands after `git commit`.

Thanks for the hint, now it works as it should.

You might want to check the script for the Knora docs.

benjamingeer:
The link you added to the README is already in the README. Did you forget to merge `develop` into this branch?

tobiasschweizer:
:-) Done

PR #136, opened 2017-02-27
build on macos, typo and issue

loicjaouen:
there was a typo (see commit)

there is a prerequisite to use `macOS Sierra`, at least, building on Yosemite, default `clang` is outdated, when trying to install `XCode`, Apple proposes xcode 8, which implies a `Serra` upgrade.
I guess it would be the same on `El Capitan`.

### DIFF
--- a/manual/building.rst
+++ b/manual/building.rst
@@ -68,7 +68,7 @@ You will need Homebrew_. Then:
     brew install cmake
     brew install doxygen
     brew install openssl
-    brew inwtall libmagic
+    brew install libmagic
     brew install gettext
     brew install nginx
     sudo chown -R $USER /usr/local/var/log/nginx/
--- a/manual/building.rst
+++ b/manual/building.rst
@@ -60,7 +60,7 @@ which are specified in the file ``Color Profile EULA.pdf``.
 Mac OS X
 ========
 
-You will need Homebrew_. Then:
+You will need Homebrew_ and at least OSX 10.11.5. Then:
 
 ::
 


benjamingeer:
Thanks! I'm using El Capitan without problems.

loicjaouen:
you are right, the current xcode 8.2.1 ships with clang 8.0.0 and its states that it's compatible with OSX 10.11.5, so the current OSX requirement is 10.11.5 (El Capitan and a half).

For Yosemite, having not switched to El Capitan by now, one has to jump from OSX to macOS Sierra.

manual/building.rst
```diff
@@ -60,7 +60,7 @@ which are specified in the file ``Color Profile EULA.pdf``.
 Mac OS X
 ========
 
-You will need Homebrew_. Then:
+You will need Homebrew_ and at least OSX 10.11.5. Then:
```
retokromer: I suggest to use `macOS`, because it’s the current official naming (also in the title).

loicjaouen: "change what you're saying, don't change what you've said" (Eels)

OSX is the official name for versions before 10.12, it was the name then:
https://web.archive.org/web/20160902012446/https://www.apple.com/osx/

And still is is according to Apple (text and pictures):
https://www.apple.com/macos/how-to-upgrade/

But `macOS` is the new mantra, so I guess we should change the title ;)

benjamingeer:
@subotic Could you check why the travis build is failing here? It looks like a problem with a command-line option of `openssl`:

```
$ openssl aes-256-cbc -K $encrypted_7deda1f68196_key -iv $encrypted_7deda1f68196_iv -in private/private.tar.enc -out private/private.tar -d
iv undefined
The command "openssl aes-256-cbc -K $encrypted_7deda1f68196_key -iv $encrypted_7deda1f68196_iv -in private/private.tar.enc -out private/private.tar -d" failed and exited with 1 during .
```

subotic:
As far as I understand it, this error means, that the file could not be decrypted, but I don't know why.

subotic:
Apparently, ``iv undefined`` means that the input vector is undefined, i.e. that it failed to read the secure variable. I've restarted an older build. Let's see if it will pass. If it does, then there might be a bug with pull requests from forks.

subotic:
There seems to be a problem with setting the environment variables, but I don't know why. Opened an issue with Travis-CI: https://github.com/travis-ci/travis-ci/issues/7412

subotic:
So, this is a feature and not a bug (see answer in https://github.com/travis-ci/travis-ci/issues/7412). Travis-CI does not decrypt secure variables when the pull-request is coming from a fork. For this to work, you would need to create a branch in the ``dhlab-basel/Sipi`` repository, and then to the PR from there.

benjamingeer:
@loicjaouen I've sent you an invitation to be a collaborator on this GitHub repository, so you should be able to make a branch for a new pull request in this repository without doing a fork.

loicjaouen:
Cool, thanks!

PR #148, opened 2017-03-24
parse Knora session id from cookie

tobiasschweizer:
This pull request fixes the parsing of the Knora session id from the cookie sent to Sipi on an image request (HTTP request header).

The session used to be numeric, now it isn's anymore. The regex had to be adjusted accordingly.

I have put the regex in a central place ad added a test for it.

I have also added a fix so that the script dir for Lua is addd before any Lua file is loaded.

Sipi now returns a content-type on login and logout (needed for jQuery, see <https://github.com/dhlab-basel/Knora/pull/473>).

### DIFF
--- a/config/sipi.fake-knora-test-config.lua
+++ b/config/sipi.fake-knora-test-config.lua
@@ -167,5 +167,10 @@ routes = {
         method = 'GET',
         route = '/test_functions',
         script = 'test_functions.lua'
+    },
+    {
+        method = 'GET',
+        route = '/test_knora_session_cookie',
+        script = 'test_knora_session_cookie.lua'
     }
 }
--- a/config/sipi.init-debug.lua
+++ b/config/sipi.init-debug.lua
@@ -19,6 +19,8 @@
 -- You should have received a copy of the GNU Affero General Public
 -- License along with Sipi.  If not, see <http://www.gnu.org/licenses/>.
 
+require "get_knora_session"
+
 -------------------------------------------------------------------------------
 -- This function is being called from sipi before the file is served
 -- Knora is called to ask for the user's permissions on the file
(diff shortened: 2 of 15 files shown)


benjamingeer:
Lovely, please merge. Please also open an issue in the Knora repository to add a Knora-Sipi integration test for this, and assign it to me.

tobiasschweizer:
Could we add a test first?

tobiasschweizer:
I mean in the Sipi tests

benjamingeer:
I really don't have time now, I'm trying to figure out whether Sipi distorts images. If you have time, please be my guest.

tobiasschweizer:
Ok, I will move the regex to a central place and write also a function that does the matching (that function will be called from within the init scripts then).

Then I will a new section to the test route we made.

But we should really think about how separating Sipi core functionality from what we need only for Knora. Next week though :-)

tobiasschweizer:
I fixed another problem which is related to this Knora fix: https://github.com/dhlab-basel/Knora/pull/473

benjamingeer:
Is this ready to review? If so, please request a review by clicking on "Reviewers" above.

tobiasschweizer:
No, it's not.

I am struggling with the `package.path` because I want to put some code in separate file.

tobiasschweizer:
Did you program a script that puts  a comment "Beautiful, thank you!" on your pull requests? :-)

benjamingeer:
> Did you program a script that puts a comment "Beautiful, thank you!" on your pull requests? :-)

Yes, beautiful, thank you! :)

tobiasschweizer:
Ok, I recompiled it from scratch, ran the tests for Sipi, then the integration tests in Knora and then the browser tests. Everything ok.

PR #170, opened 2017-06-18
Improvements

### LINKED ISSUE #168: Rename config options docroute and docroot to be less confusing
The config options `docroute` and `docroot` sound the same, let's rename them.

lrosenth:
### Summary

@benjamingeer @tobiasschweizer This new version incorporates several bug fixes and improvements:

- error message if script init file is not found
- —query (-x) option for getting all metadata of a file
- Adding subdirs 'A' - 'Z' to image directory. The number of subdirs is controlled by a config file variable ``subdir_levels``.
- Fixed bug with ICC profile (crash)if one of the J2K reference files
- extended documentation of Sipi-Lua functions
- Since the config variables „docroot“ and „docroute“ are easy to mix, „docroute“ has bee changed to „wwwroute“
- if the format specificer „-f“ or „—format“ is missing, SIPI tries to determine the output file format using the extension of the filename. Supported are „jpx“, „jp2“, „tif“, „tiff“, „jpg“, „jpeg“ and „png“.
- Removed restriction for XMP since new version of Exiv2 is used (V0.26) that should make XMP thread safe

Attention: there is still one of the J2K reference files which SIPI cannot convert. It has to to with kakadu and tiling...

### Issues

 - resolves #168


lrosenth:
travis: compiling Sipi seems to work – the error is in running the tests... (as far as I get it)

subotic:
yes, this is what I also understand. you can run the tests with `make check`

src/sipi.cpp
```diff
@@ -564,11 +619,32 @@ int main(int argc, char *argv[]) {
...
         }
+        else {
+            //
+            // there is no format option given – we try to determine the format
+            // from the output name extension
+            //
+            size_t pos = outfname.rfind(".");
+            if (pos != std::string::npos) {
+                std::string ext = outfname.substr(pos + 1);
+                if ((ext == "jpx") || (ext == "jp2")) {
```
benjamingeer: What happens if `--format` is not given, and we get an unrecognised file extension, or if there is no file extension?

lrosenth: Solved. If the file extension is missing or unknown (and no --format clause is given), an error message is emitted and the command line conversion terminates.

tobiasschweizer:
@lrosenth How do we make sure that new directory organization is compatible with a running Sipi instance that has all the files in one directory? How would one switch to the new directory structure with a lot of files?

PR #178, opened 2017-06-26
fix (config): add subdir exludes to docker config

subotic:
### Summary

This should have been part of the last PR, but I forgot to update the docker configs.

- add missing exclude dirs variable to docker configuration


subotic:
Thanks :-)

PR #190, opened 2017-07-04
Add 'unzip' to Ubuntu 16.04 prerequisites

schwemmer:
In order to unzip Kakadu we also need unzip installed. Add `sudo apt-get install unzip` to prerequisites for Ubuntu 16.04


subotic:
Thanks :-)

PR #212, opened 2017-09-01
accept upper-case file extensions for mime type check

tobiasschweizer:
When creating an image, we check for the consistency of the mime type indicated by the user, the original file extension and the actual mime type (determined by libmagic).

In order to check if the file extension matches a mime type, we look it up in a map whose keys are all lower-case.

This PR converts the file extension to lower-case so also upper-case extensions are found in the map.


tobiasschweizer:
@lrosenth Why does `::tolower` not have a prefix `std`?

PR #214, opened 2017-09-21
docs (use optimization): compile Sipi with optimization level 3

tobiasschweizer:
added to docs how to compile with optimisation level 3


tobiasschweizer:
Für wen?

PR #244, opened 2018-06-26
build (docker): make smaller image

subotic:
### Summary

- add linker flags: `-static` `-static-libgcc` `-static-libstdc++`
- add second build stage only containing the binary and the most necessary directories and files


subotic:
with the move to Bazel, we will get smaller images.

PR #323, opened 2020-04-20
builld: add makefile targets for compiling and running sipi inside docker

subotic:


kilchenmann:
Seems to work. There was just one error in `make run` because image folder is missing. Probably we should add a comment to the Readme?

```zsh
Error at [/sipi/src/SipiFilenameHash.cpp: 85] (system error: No such file or directory): Couldn't read directory content! Path: ./images
```

subotic:
I can create it before running sipi in the makefile. I will add it tomorrow.

kilchenmann:
> I can create it before running sipi in the makefile.

...if it doesn't exist? But I mean, it's only a detail.

Beside that it works great.

kilchenmann:
@subotic I saw now, that I have uncommitted changes (or untracked files) in the repository. All in `/test/_test_data/images/unit/...`. Should we push them or adding them to `.gitignore`?

subotic:
@kilchenmann So now the images folder is created if not there and the test images are ignored.

subotic:
great, thanks :-)

PR #331, opened 2020-09-23
Fix/issues fixer

lrosenth:
Latest version... Should fix knora.json problem


benjamingeer (approved):
I can't review this properly without a user story and explanation for each change, and it would be easier to review if each user story was implemented in a separate, smaller pull request. But let's try it. :)

benjamingeer (changes requested):
Sorry, I forgot something: does this PR fix the version number and commit ID printed on startup?

shttps/Parsing.cpp
```diff
@@ -96,19 +95,21 @@ namespace shttps {
...
                 {"aifc",    {"audio/x-aiff"}},
                 {"mid",     {"audio/x-midi"}},
                 {"midi",    {"audio/x-midi"}},
                 {"mp2",     {"audio/x-mpeg"}},
-                {"wrl",     {"model/vrml",                                                        "x-world/x-vrml"}},
+                {"wrl",     {"model/vrml", "x-world/x-vrml"}},
                 {"ics",     {"text/calendar"}},
-                {"csv",     {"text/comma-separated-values",                                       "text/plain"}},
-                {"tsv",     {"text/tab-separated-values",                                         "text/plain"}},
+                {"csv",     {"text/comma-separated-values", "text/plain", "application/vnd.ms-excel",
```
benjamingeer, line 110: Currently the `/knora.json` route returns `text/comma-separated-values`  for a CSV file, but this MIME type is unregistered. The registered MIME type for CSV is `text/csv`: https://www.rfc-editor.org/rfc/rfc4180.html

lrosenth:
@benjamingeer  The csv-mimetype is returned as "text/csv" in the JSON variable "internalMimeType"...
Should now be ready for merge and deployement...

benjamingeer:
Does this PR fix the version number and commit ID printed on startup (https://dasch.myjetbrains.com/youtrack/issue/DSP-687)?

It would be great to have this automatically updated. CMake can generate a source file with the current Git commit hash: https://stackoverflow.com/questions/1435953/how-can-i-pass-git-sha1-to-compiler-as-definition-using-cmake/4318642#4318642

subotic:
@benjamingeer I don't think so

> Does this PR fix the version number and commit ID printed on startup (https://dasch.myjetbrains.com/youtrack/issue/DSP-687)?

I don't think so. There is another PR open to solve this. The checksum printed at startup comes already from git and cmake. The only thing missing is the tag. It is next on my list.

benjamingeer (approved):
> There is another PR open to solve this. The checksum printed at startup comes already from git and cmake. The only thing missing is the tag. It is next on my list.

OK then let's merge this now and make a new Sipi release from it.

PR #334, opened 2020-10-26
fix for upscaling IIIF V3

lrosenth:
Under certain circumstances upscaling an image using the IIIF V3 syntax made SIPI crash


subotic (approved):
LGTM :-) There are a few strange changes shown, which were added in a previous PR. Also, your commit history is a bit long.

PR #335, opened 2020-11-05
Bugfix: SIPI crashed for some invalid IIIF url's

lrosenth:
If Sipi received invalid IIIF url's with missing parts, Sipi could crash.
The bug is fixed and assorted tests for this behaviour have been added.


subotic (approved):
Great, thanks!

PR #340, opened 2021-01-21
fix: parse url crash

lrosenth:
Issue: DSP-1247

Should correct fix which make SIPI crash in the dsp-api stack if the image size was less than 256 in one dimension. Sipi newly uses no tiles for small images (which makes sense over all)


subotic (approved):
Great! :-) I even understood everything :-)

PR #344, opened 2021-02-01
add: sidecar support

lrosenth:
Added support to process sidecar file for "knora.json".


subotic (approved):
cool, thanks! looks good.

PR #373, opened 2022-09-08
add support for ODD and RNG file types (DEV-1271)

mpro7:
This PR should add support for ODD and RNG file types.


subotic:
deprecated by #374

PR #381, opened 2023-02-02
feat: Add triggering of preflight script for non-image file types (DEV-1664)

subotic:
- DEV-1664 Add triggering preflight script for non image file types


lrosenth (approved):
All test pass on apple silicon native, intel nativ, linux docker

PR #383, opened 2023-02-20
fix: incorrect error message

BalduinLandolt:
the error message does not reflect the lua function that has actually been called


subotic (approved):
Danke 👍

PR #384, opened 2023-02-27
Wip/mimetypes

lrosenth:
Added support for upload/downlaod of different mime types, especially *.odd files (which return as "best" mime type "text/xml"...

See "upload.lua" in the testing suite.
In addition the Lua-function "img:gps()" returns a table of available GPS data from the EXIF-header. Also there see the testing for code example.


subotic (approved):
LGTM, thanks 👍

PR #397, opened 2023-11-02
refactor(sentry): add per request session tracking

subotic:


shttps/Server.cpp
```diff
@@ -721,6 +774,11 @@ namespace shttps {
     }
     //=========================================================================
 
+    // process request is one thread that is waiting for incoming requests
+    // and processes them. It is started by the main thread and will run
+    // until the server is stopped.
+    // This is in addition to the thread pool that is used to process the
+    // lua scripts.
```
BalduinLandolt, line 781: does this mean that all non-lua logic is single threaded?

BalduinLandolt (approved):

shttps/Server.cpp
subotic: no, everything is multithreaded and this comment is a bit misleading. There is only one thread pool used by all routes. I will correct it.

subotic: @BalduinLandolt can you maybe take again a look at the changes?

subotic: > does this mean that all non-lua logic is single threaded?

also please note the new naming of the methods. Before we had `process_request` and `processRequest`. I hope the new naming is a bit more clearer what is happening (at least what I think that is happening).

shttps/Server.cpp
```diff
@@ -721,7 +724,9 @@ namespace shttps {
     }
     //=========================================================================
 
-    static void *process_request(void *arg) {
+    // The socket_request_processor is added to each tread and initiates handling of requests
```
BalduinLandolt, line 727:
```suggestion
    // The socket_request_processor is added to each thread and initiates handling of requests
```

BalduinLandolt (approved):
looks good/better, just found one typo

shttps/Server.cpp
subotic: thanks, changed now.

PR #401, opened 2023-11-22
refactor(sentry): use crashpad as sentry backend

subotic:
Part of DEV-2988

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,18 +41,6 @@ docker-build: ## build and publish Sipi Docker image locally
 		--load \
 		.
 
-.PHONY: docker-build-debug
-docker-build-debug: ## build and publish Sipi Docker image locally with debugging enabled
-	docker buildx build \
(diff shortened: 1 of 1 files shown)


src/sipi.cpp
```diff
@@ -1000,10 +995,6 @@ int main(int argc, char *argv[]) {
 
         try {
             std::cout << std::endl << "Ivan was here" << std::endl;
```
BalduinLandolt, line 997: Is this line still needed? :P

BalduinLandolt (approved):
I understand very little of what's going on here, but from what I see, it makes sense

src/sipi.cpp
subotic: I like it ;-) It is also used in some queries in Infra.

subotic: Sadly, the only way to figure out if Sipi restarted.

PR #404, opened 2024-01-22
doc: Sipi only allows for `restrict` and not `restricted` in the permission table in lua scripts

seakayone:


mpro7 (approved):
Thanks! Why does it failing on updating MD file?

subotic (approved):

PR #410, opened 2024-02-14
fix: support for grayscale jpegs

subotic:
related to DEV-3271


lrosenth (approved):
Looks perfect for the problem with the gray valued JPEG!

subotic:
> Looks perfect for the problem with the gray valued JPEG!

great, thanks!

PR #412, opened 2024-03-05
feat: gray values for transparent parts of image when returning a JPEG

subotic:


src/formats/SipiIOJpeg.cpp
```diff
@@ -1131,8 +1010,14 @@ namespace Sipi {
         // we have to check if the image has an alpha channel (not supported by JPEG). If
         // so, we remove it!
         //
-        if ((img->getNc() > 3) && (img->getNalpha() > 0)) { // we have an alpha channel and possibly a CMYK image
-            img->removeExtraSamples();
+        const int number_of_alpha_channels = img->getNalpha();
+        const int number_of_channels = img->getNc();
+        bool three_or_more_channels = number_of_channels > 3;
```
siers, line 1015: shouldn't this be "more than three"?

include/SipiImage.h
```diff
@@ -178,36 +181,27 @@ class SipiImageError final : public std::exception {
...
         PhotometricInterpretation photo;    //!< Image type, that is the meaning of the channels
-        byte *pixels;   //!< Pointer to block of memory holding the pixels
-        std::shared_ptr<SipiXmp> xmp;   //!< Pointer to instance SipiXmp class (\ref SipiXmp), or NULL
-        std::shared_ptr<SipiIcc> icc;   //!< Pointer to instance of SipiIcc class (\ref SipiIcc), or NULL
-        std::shared_ptr<SipiIptc> iptc; //!< Pointer to instance of SipiIptc class (\ref SipiIptc), or NULL
-        std::shared_ptr<SipiExif> exif; //!< Pointer to instance of SipiExif class (\ref SipiExif), or NULL
-        SipiEssentials emdata; //!< Metadata to be stored in file header
-        shttps::Connection *conobj; //!< Pointer to HTTP connection
-        SkipMetadata skip_metadata; //!< If true, all metadata is stripped off
+        byte *pixels;                       //!< Pointer to block of memory holding the pixels (allways in big-endian format if interpreted as 16 bit/sample)
```
siers, line 197: "allways"

src/SipiImage.cpp
```diff
@@ -508,68 +508,131 @@ namespace Sipi {
...
+                es.clear();
+            }
+
+            // TODO: figure out when this can happen. Maybe two channels with alpha is not allowed or even possible?
+            if (has_three_channels) {
+                std::string msg = "Cannot remove component: nc=" + std::to_string(nc) + " chan=" + std::to_string(channel);
+                throw SipiImageError(__file__, __LINE__, msg);
+            }
+
+            // TODO: figure out when this can happen and if this can/should be cought earlier
```
siers, line 540: caught

siers (approved):

src/formats/SipiIOJpeg.cpp
subotic: yes, you are right 👍

src/SipiImage.cpp
```diff
@@ -508,68 +508,129 @@ namespace Sipi {
...
+        const bool has_two_or_less_channels = nc < 3;
+        const bool has_three_channels = nc == 3;
+
+        // Assumtion: An image with two or less channels cannot have extra samples
+        assert(has_two_or_less_channels && has_removable_extra_samples);
+
+        if (has_removable_extra_samples) {
+
+            // cleanup the extra samples
+            // TODO: figure out why this can even happen
```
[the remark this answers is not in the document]
subotic: It is a reminder to my future self. The only reason why it is a todo is that clion colors it differently so that I can easily spot it. Ideally there should be no TODOs but only actionable issues on Linear. Unfortunately, at this point, I'm still figuring out what Sipi is doing. As soon as I know, it will be turned into an issue.

src/SipiImage.cpp
```diff
@@ -508,68 +508,129 @@ namespace Sipi {
...
-            } else {
-                std::string msg = "Cannot remove component: nc=" + std::to_string(nc) + " chan=" + std::to_string(chan);
+        const bool has_removable_extra_samples = !es.empty();
+        const bool has_two_or_less_channels = nc < 3;
+        const bool has_three_channels = nc == 3;
+
+        // Assumtion: An image with two or less channels cannot have extra samples
+        assert(has_two_or_less_channels && has_removable_extra_samples);
+
+        if (has_removable_extra_samples) {
```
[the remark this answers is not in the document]
subotic: yes, the proposed inspect thingy. Unfortunately this is still not part of the language as of c++23. Anyway, my hope is that after a thorough refactoring, this horrible if-tree will not be needed anymore. It simply tries to figure out the image characteristics. Those could, in my opinion, be better modeled, so that it is clear what image I have in hand and what operations are allowed.

subotic:
@siers and @BalduinLandolt thanks for the review.

PR #449, opened 2024-07-22
fix: update ijg jpeg lib to v9f, partially fixes DEV-3474

siers:
As far as I understand, the core of the issue is that the ijg library used not to correctly recognize when the 4 channel colorspace is actually YCCA, which is likely (but I can't say with 100% certainty) fixed in v9e: https://github.com/libjpeg-turbo/ijg/commit/ff9491d4e0a06542ae82367183e4548664705d80#diff-71516311e7ecc262ab53731fc7ec0ea1c6d50ff9367bdea67804f9af5c9015e5L165-R180

I discovered where the bug lies after checking that the values of a image of just cyan (100%, 0%, 0%, 0%) are 0, 255, 255, 255, so it could not have been the fault of the image profile convertor (LittleCMS, in this case. In addition, I tried setting all the parameters for LCMS manually and tested it with a byte array, which was all correct.)

Knowing the root cause, additional reports of such problems can be found, because this feature space is probably not extensively tested.

With the ijg lib updated to v9f, the original file attached in the issue is getting converted correctly, although files converted with imagemagick still remain black.

### DIFF
--- a/ext/jpeg/CMakeLists.txt
+++ b/ext/jpeg/CMakeLists.txt
@@ -9,13 +9,13 @@ set(timestamp_policy DOWNLOAD_EXTRACT_TIMESTAMP OLD)
 
 ExternalProject_Add(project_jpeg
         INSTALL_DIR ${CMAKE_BINARY_DIR}/local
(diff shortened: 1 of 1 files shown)


seakayone (approved):

subotic (approved):
Cool, thanks!

PR #451, opened 2024-08-05
feat: Allow YCbCr/RGB autoconvert, update libtiff (DEV-3863)

siers:
* patch removed, because it now seems to be fixed in the upstream
* test-e2e makefile command fixed
* unit test with the buggy picture added


subotic (approved):
Thanks!

PR #453, opened 2024-09-02
feat: respect X-Forwarded-Proto (DEV-3499)

siers:
`X-Forwarded-Proto` is now being respected in both `info.json` and `knora.json`. Repetition slightly reduced. Both are tested in `test/e2e`.

See also:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto


subotic (approved):
Very good. Thank you!

PR #455, opened 2024-09-23
fix: Fix planar BigTIFF ingestion (DEV-3384)

siers:
Since the files that are to be tested range in the multiple gigabytes (when uncompressed) and will take a lot of space and time when testing, they have not been added to the test suite. Accordingly, the fix will be tested on the dev env.

As before, when ingesting a tiled BigTiff, an error noting the incompatibility will be issued.


subotic (approved):
👍

PR #456, opened 2024-09-24
feat: Support IIIF Image's API HEAD requests (DEV-4072)

siers:
HEAD requests should return the status code without the body content. E2E Tests included.


src/SipiHttpServer.cpp
```diff
@@ -1191,6 +1191,7 @@ static void serve_iiif(Connection &conn_obj,
   const std::string &uri,
   std::vector<std::string> params)
 {
+  auto headRequest = conn_obj.method() != Connection::HEAD;
```
subotic, line 1194: Maybe better to call it `not_head_request`?

subotic (approved):

PR #459, opened 2024-12-17
fix: Detect corrupted files, exit on exception (DEV-4457)

siers:
This corruption warning is non-fatal, so we have to check if the error code is set to a specific JWRN after jpeg_read_scanlines. Test added that an exception is created in img.read.

---

References:

https://github.com/mozilla/mozjpeg/blob/6c9f0897afa1c2738d7222a0a9ab49e8b536a267/jerror.h#L195
```
JMESSAGE(JWRN_HUFF_BAD_CODE, "Corrupt JPEG data: bad Huffman code")
```
https://github.com/mozilla/mozjpeg/blob/6c9f0897afa1c2738d7222a0a9ab49e8b536a267/jerror.h#L280-L283
```
/* Nonfatal errors (we can keep going, but the data is probably corrupt) */
#define WARNMS(cinfo,code)  \
  ((cinfo)->err->msg_code = (code), \
   (*(cinfo)->err->emit_message) ((j_common_ptr) (cinfo), -1))
```
https://github.com/mozilla/mozjpeg/blob/6c9f0897afa1c2738d7222a0a9ab49e8b536a267/jdphuff.c#L551-L552
```
        if (s) {
          if (s != 1)           /* size of new coef should always be 1 */
            WARNMS(cinfo, JWRN_HUFF_BAD_CODE);
          CHECK_BIT_BUFFER(br_state, 1, goto undoit);
```
https://github.com/mozilla/mozjpeg/blob/6c9f0897afa1c2738d7222a0a9ab49e8b536a267/jdhuff.c#L463-L466
```
  if (l > 16) {
    WARNMS(state->cinfo, JWRN_HUFF_BAD_CODE);
    return 0;                   /* fake a zero as the safest result */
  }
```


test/_test_data/images/unit/jpegCorrupt.jpg
jnussbaum, line 1: I think there's a misunderstanding: This file is the output of SIPI, not the input. The input file is the one where the lower part is entirely grey, without structure.

src/sipi.cpp
```diff
@@ -862,6 +862,7 @@ int main(int argc, char *argv[])
       }
     } catch (Sipi::SipiImageError &err) {
       std::cerr << err << std::endl;
+      exit(1);
```
jnussbaum, line 865: Hmmm, does that mean that in the past, all `SipiImageError` resulted in return code 0, but now, they will all result in return code 1? 

That would not be the expected behaviour. Because often, an image is only slightly corrupted, but perfectly readable by most viewers. For example, the ca. 66'000 images of LHTT produced ca. 14'000 error logs. But most of them are negligible, because there is no real defect in the image. We'd like that SIPI only aborts with non-0 in these very specific cases:

- `Corrupt JPEG data: premature end of data segment` (https://linear.app/dasch/issue/DEV-4480)
- `Corrupt JPEG data: bad Huffman code` (https://linear.app/dasch/issue/DEV-4457)

test/_test_data/images/unit/jpegCorrupt.jpg
siers: Isn't `47fdDI7XARj-dD5pt3RUc2e.jpg.orig` the original JPG file (coming from the zip in the ticket)?

jnussbaum: Exactly, but this is not `47fdDI7XARj-dD5pt3RUc2e.jpg.orig`. You took `47fdDI7XARj-dD5pt3RUc2e.jpx` by accident.

siers: Unziped the zip again, just to check there were no accidental overwrites/copies:
```
% md5sum test/_test_data/images/unit/jpegCorrupt.jpg
894f7f58c8adc76bb430f0953814df13  test/_test_data/images/unit/jpegCorrupt.jpg
% md5sum /tmp/sand-pbr/*
e1f5b859cfb4cebb7078a55b390339e3  /tmp/sand-pbr/47fdDI7XARj-dD5pt3RUc2e.info
894f7f58c8adc76bb430f0953814df13  /tmp/sand-pbr/47fdDI7XARj-dD5pt3RUc2e.jpg
894f7f58c8adc76bb430f0953814df13  /tmp/sand-pbr/47fdDI7XARj-dD5pt3RUc2e.jpg.orig
8f827e5ccd67a3b088dab516f27bae55  /tmp/sand-pbr/47fdDI7XARj-dD5pt3RUc2e.jpx
```
It seems to be the `.jpg.orig`. Are you sure? What have I overlooked?

src/sipi.cpp
siers: Very odd, but it was also my intention to check whether this new behavior breaks the build. Unfortunately the `main` build is broken. 🤦

Anyhow, I'll restrict it to the behavior you've described.

test/_test_data/images/unit/jpegCorrupt.jpg
jnussbaum: Okay, you're right. 

I was confused because in Mac's "Preview" on my machine, the `jpg.orig` is grey in the lower part, while the `jpx` has some structure in the lower part. 

In GitHub's browser view, `test/_test_data/images/unit/jpegCorrupt.jpg` looks like the `jpx` on my machine.

But if the checksum is correct, then you're of course right.

src/sipi.cpp
siers: Addressed in the last commit, after I resolved the build issues in another PR.

jnussbaum: @siers In this notion page, we discussed that we should probably not modify the SIPI behaviour at all -> https://www.notion.so/SIPI-error-logs-1748946b7d4080618e91cf0527b9d7a7?d=1748946b7d408031a344001cf88f05f8&pvs=4#1748946b7d408058825ac23f05bcfc38

Have you seen the discussion there?

Sorry that this comes only now, after you have already invested time. We should have spent more time on thinking about the issue, before creating a Linear ticket. 

I suggest that you don't spend any more time on this PR, before we have decided what to do.

PR #465, opened 2025-01-23
feat: return 404 for missing files (INFRA-735)

siers:
Technically, the underlying function checks for file accessibility, but an internal server error would be inappropriate even then.

### DIFF
--- a/test/e2e/test_02_server.py
+++ b/test/e2e/test_02_server.py
@@ -99,7 +99,7 @@ class TestServer:
             "/knora/DenyLeaves.jpg/full/max/0/default.jpg", 401)
 
     def test_not_found(self, manager):
-        """return 401 Unauthorized if the user does not have permission to see the image"""
+        """return 404 Not Found if the file is missing. Sipi will try and fail to find this file in the image directory."""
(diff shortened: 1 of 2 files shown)


test/e2e/test_02_server.py
```diff
@@ -98,6 +98,11 @@ def test_deny(self, manager):
         manager.expect_status_code(
             "/knora/DenyLeaves.jpg/full/max/0/default.jpg", 401)
 
+    def test_not_found(self, manager):
+        """return 401 Unauthorized if the user does not have permission to see the image"""
+        manager.expect_status_code(
+            "/file-should-be-missing-123", 404)
```
seakayone, line 104: *question* what should the test actually check? 

I assume the description should talk about a 404 and not an unauthorized...

siers: Mindless focus, sorry. Will fix in a second.

siers: Fixed.

seakayone (approved):

SamuelBoerlin:
Thanks for the fix @siers ! :)

PR #470, opened 2025-03-18
fix: Logger.cpp in LuaServer.cpp, remove syslog, missing JSON quotes

siers:
* Start using Logger.cpp in LuaServer
    * add the missing log levels LuaServer used
    * thereby removing remaining syslog uses
* Small test suite for Logger.cpp, fix missing quotes
* Remove log levels for now (will update, if poses problems)
* Includes some autoformatter changes

### DIFF
--- a/shttps/Server.cpp
+++ b/shttps/Server.cpp
@@ -524,11 +524,6 @@ Server::Server(int port_p,
     ll = LL_ERR;
   }
 
-  // TODO: setlogmask disabled entirely for now
-  // openlog(loggername, LOG_CONS | LOG_PERROR, LOG_DAEMON);
-  // openlog(loggername, LOG_PERROR, LOG_DAEMON);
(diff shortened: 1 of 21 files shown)


shttps/Server.cpp
```diff
@@ -506,28 +505,29 @@ Server::Server(int port_p,
...
   } else {
-    ll = LOG_ERR;
+    ll = LL_ERR;
   }
 
+  // TODO: setlogmask disabled entirely for now
   // openlog(loggername, LOG_CONS | LOG_PERROR, LOG_DAEMON);
-  openlog(loggername, LOG_PERROR, LOG_DAEMON);
-  setlogmask(LOG_UPTO(ll));
+  // openlog(loggername, LOG_PERROR, LOG_DAEMON);
```
subotic, line 529: Do we need the old syslog things? Either it works now or it doesn't.

subotic (approved):

PR #478, opened 2025-07-09
fix: TIFF tiled reading bug

siers:
* `memcpy` in `read_tiled_data` has been rewritten as a loop with guards completely eliminating out of bounds access.
* Bug discovered: Reading from smaller dimensions in the TIFF pyramid has been commented out, because it introduces a bug together with a region `0,0,x,x` and a size `pct:50`.
* Bug discovered: `pct:50` with a rectangular region won't produce a rectangular output, which is not according to the spec.


subotic (approved):
Thanks!

PR #587, opened 2026-04-29
feat(icc): deterministic ICC creation date for bit-exact tests (DEV-6333)

subotic:
[DEV-6333](https://linear.app/dasch/issue/DEV-6333)

## Summary

- `SipiIcc::iccBytes()` now overwrites bytes 24-35 (ICC creation date) and zeros bytes 84-99 (Profile ID) when `SOURCE_DATE_EPOCH` is set. Production never sets it, so deployed binaries are bit-identical.
- CMake injects `SOURCE_DATE_EPOCH=946684800` for `sipi.approvaltests` only, so JPEG / PNG / JP2 outputs become byte-deterministic under ctest.
- Approval baseline grows from 5 TIFF tests (PR 0) to 15 tests covering every codec and the lcms2 ICC emit path.

## Changes

**Helper + chokepoint integration**
- `src/metadata/SipiIcc.cpp` — adds `Sipi::detail::apply_icc_header_normalization` (pure byte mutation) and `read_source_date_epoch` (cached env read, .cpp-private). One-line call inside `iccBytes(unsigned int&)`. Also fixes a pre-existing leak on the second `cmsSaveProfileToMem` failure path.
- `include/metadata/SipiIccDetail.h` — new internal header so unit tests can drive the pure function with explicit epoch values.
- `include/metadata/SipiIcc.h` — Doxygen comment + `[[nodiscard]]` on both `iccBytes()` overloads.

**Unit tests**
- `test/unit/sipiicc/` — 8-case GoogleTest suite covering: nullopt → no-op, epoch=Y2K → expected date bytes, Profile ID zeroed, non-zero ID scrubbed, truncated buffer no-op, idempotency, only-documented-byte-ranges-touched, out-of-range-year → no-op (covers `gmtime_r` returning NULL on 32-bit `time_t` post-2038).

**Approval-test extension**
- `test/approval/CMakeLists.txt` — `set_tests_properties(... ENVIRONMENT SOURCE_DATE_EPOCH=946684800 ...)`.
- `test/approval/image_encode_baseline_test.cpp` — 10 new tests: `JpegFullToJpegDownscaled`, `JpegFullToPng`, `J2kRegionToTiff`, `J2kRegionToJpeg`, `J2kRegionToJp2` (Goal-3 Kakadu-encode determinism check), `CmykTiffToJpeg`, `CielabTiffToJpeg`, `JpegRotatedToJpeg`, `PngRoundTrip`, `TiffToPng`. Top-of-file commentary updated.
- 11 new LFS-tracked goldens; 2 existing TIFF goldens re-captured (their input ICCs round-trip through lcms2 and pick up the normalized date — see "Deviation from plan" below).
- New PNG fixture `test/_test_data/images/unit/sample_with_icc.png` generated by SIPI itself; `gray_with_icc_another.jpg` (grayscale + sRGB-ICC mismatch — tripped lcms2's input colour-space check) replaced with `MaoriFigureWatermark.jpg`.

**Documentation**
- `CLAUDE.md` — "ICC determinism invariant" callout in the build invariants section.
- `docs/src/development/testing-strategy.md` — ICC determinism gate documented in Layer 2.
- `docs/src/development/developing.md` — direct-binary run instructions for `sipi.approvaltests`; complete unit-test directory list (now includes `sipiicc`).
- `docs/adr/0002-icc-profile-determinism-test-only.md` — new ADR.
- `UBIQUITOUS_LANGUAGE.md` — adds *ICC normalization* and *Reproducibility flag*.
- `test/approval/CHANGELOG.approval.md` — env-var gate explanation + audit-log entry for the 2 re-captured TIFF goldens.

## Deviation from plan

Plan AC item *"Existing TIFF goldens from PR 0 unchanged after the fix (`git diff test/approval/approval_tests/Image*.approved.tif` empty)"* turns out to be infeasible:

- `TiffRegionRoundTrip` (input `lena512.tif`) carries a *Generic Gray Gamma 2.2* ICC.
- `CmykTiffDownscaled` (input `cmyk.tif`) carries a *U.S. Web Coated (SWOP)* ICC.

Both round-trip through `cmsSaveProfileToMem`, so under `SOURCE_DATE_EPOCH` their date bytes get normalized just like a fresh lcms2 profile would. `cmp -l` confirms the diff is exactly the 12-byte `dateTimeNumber` field per embedded ICC profile — surgical, not real content drift. The other 3 PR 0 goldens (no embedded ICC) are unchanged byte-for-byte.

The production-equivalent invariant *does* hold: without `SOURCE_DATE_EPOCH`, all 5 PR 0 TIFF goldens still pass against the binary built from this branch. The plan's intent — "the helper is a no-op in production" — is preserved; the AC's wording was just over-specific.

## Test Plan

- [x] `just nix-build` (`.#dev` derivation) succeeds; sandbox `checkPhase` runs unit + approval tests.
- [x] `cd build/test/approval && SOURCE_DATE_EPOCH=946684800 ./sipi.approvaltests --gtest_filter='ImageEncodeBaseline.*'` — 15/15 PASS, twice in a row, zero `.received.*` files.
- [x] `cd build/test/unit/sipiicc && ./sipi_icc_normalize_test` — 8/8 PASS.
- [x] Unset env var run: 3 of 5 existing PR 0 TIFF goldens still PASS (the 2 ICC-carrying ones intentionally drift); new ICC-touching tests intentionally drift (documented in CHANGELOG).
- [x] 4-reviewer pass (cpp / security / simplicity / consistency); review-driven follow-ups in last 4 commits.
- [ ] CI (Linux): runs in PR.

## Screenshots

n/a — no UI changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -31,6 +31,8 @@ If a change can't be verified on a platform locally, say so explicitly and gate 
 
 **First-time setup for the dev-shell inner loop:** run `just kakadu-fetch` once to download the proprietary Kakadu archive from `dsp-ci-assets` into `vendor/`. Nix builds (including `just nix-docker-build`) fetch Kakadu directly via the flake's FOD (no `vendor/` step). Requires `gh auth login` and `dasch-swiss` org membership. See [`docs/src/development/kakadu.md`](docs/src/development/kakadu.md).
 
+**ICC determinism invariant:** [`SipiIcc::iccBytes()`](src/metadata/SipiIcc.cpp) is the single chokepoint that converts an `cmsHPROFILE` into raw bytes for codec consumption — every TIFF, JPEG, PNG, and JP2 emission funnels through it. Any new format handler must route through `iccBytes()`; bypassing it (calling `cmsSaveProfileToMem` directly) breaks the approval-test gate. Approval tests run with `SOURCE_DATE_EPOCH` injected by CMake (see [`test/approval/CMakeLists.txt`](test/approval/CMakeLists.txt)) so the wall-clock-stamped ICC creation date is overwritten with a fixed value and goldens stay byte-deterministic. Production never sets the env var; deployed binaries continue to embed wall-clock-stamped ICC headers. See [`docs/adr/0002-icc-profile-determinism-test-only.md`](docs/adr/0002-icc-profile-determinism-test-only.md).
+
 ### Quick Reference
 
 ```bash
--- a/UBIQUITOUS_LANGUAGE.md
+++ b/UBIQUITOUS_LANGUAGE.md
@@ -32,6 +32,8 @@ Canonical terminology for the SIPI repository. SIPI is a IIIF Image API 3.0 serv
 | --- | --- | --- |
 | **Format handler** | A SipiIO subclass that adapts a codec to SIPI's read/write contract (SipiIOJ2k, SipiIOTiff, SipiIOPng, SipiIOJpeg). | IO backend, format driver |
 | **Codec** | A third-party library that performs the actual encode/decode (kakadu, libtiff, libpng, libjpeg, libwebp). A format handler *uses* a codec. | library, backend |
+| **ICC normalization** | The byte-level rewrite of bytes 24-35 (creation date) and 84-99 (Profile ID) inside `SipiIcc::iccBytes()`, gated by the *Reproducibility flag*. Test-only — production iccBytes() is the identity. | ICC scrubbing, ICC stripping (those imply removing profiles, not normalizing them) |
+| **Reproducibility flag** | The `SOURCE_DATE_EPOCH` environment variable. When set, every ICC profile emitted by SipiIcc::iccBytes() has its creation date overwritten with the supplied epoch and its Profile ID zeroed; codec-bound emissions become byte-deterministic. CMake injects it for `sipi.approvaltests` only. | deterministic mode, test-only mode (the env var is the contract; "modes" obscure that) |
 
 ## Metadata and preservation
 
--- a/docs/adr/0002-icc-profile-determinism-test-only.md
+++ b/docs/adr/0002-icc-profile-determinism-test-only.md
@@ -1,0 +1,7 @@
+# ICC profile creation date is deterministic in tests, wall-clock in production
+
+`SipiIcc::iccBytes()` is the single chokepoint that converts `cmsHPROFILE` to bytes for codec consumption. When `SOURCE_DATE_EPOCH` is set in the environment, it overwrites bytes 24-35 (ICC creation date) with the supplied epoch and zeros bytes 84-99 (Profile ID). When unset — the production default — it returns lcms2's wall-clock-stamped bytes verbatim.
+
+We accept this split because byte-for-byte approval testing of JPEG / PNG / JP2 outputs is otherwise impossible (lcms2's `cmsCreateProfilePlaceholder` calls `time(NULL)` and lcms2 upstream rejected `SOURCE_DATE_EPOCH` support — Debian #814883, Little-CMS #71). Production keeps wall-clock timestamps so downstream consumers reading `cmsGetHeaderCreationDateTime` see real creation times, matching every other IIIF server.
+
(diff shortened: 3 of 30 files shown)


subotic:
Review pass complete. 4 follow-up commits address the agent reviews:

**Robustness (cpp + security)** — `95ef520 fix(icc): bail on gmtime_r failure / out-of-range year`
- `gmtime_r` return value was discarded; on POSIX it can return NULL with `errno=EOVERFLOW` for an out-of-range `time_t`. Old code then read tm{} zero-initialized and emitted year 1900 silently. Now: bail (no-op).
- `static_cast<uint16_t>(tm_year + 1900)` could silently wrap for years > 65535. Now: bounds-check `[0, 65535]`, no-op outside.
- New `EpochProducingOutOfRangeYearIsNoOp` test.

**Simplicity (code-simplicity)** — `dda2443 refactor(icc): unexport read_source_date_epoch + drop redundant tests`
- `read_source_date_epoch` was declared in `SipiIccDetail.h` but no test ever called it (the magic-static cache makes it untestable in isolation). Moved back to the .cpp anonymous namespace; detail header now exposes only what tests need.
- Dropped `EmptyBufferIsNoOp` (subsumed by `TruncatedBufferIsNoOp`) and `NulloptEpochMatchesMalformedEnvBehaviour` (duplicate of `NoEpochLeavesBufferUnchanged`). 8 unit tests, same coverage.

**Documentation (consistency)** — `8a722b7 docs(icc): log re-approved goldens + complete unit-test directory`
- Added the missing re-approval audit row in `CHANGELOG.approval.md` for `TiffRegionRoundTrip` + `CmykTiffDownscaled` (per the CHANGELOG's own step 4).
- Filled in `developing.md` unit-test directory list (was missing `cache`, `decode_dims`, `handlers`, `memory_budget`, `ratelimiter`, and the new `sipiicc`).
- `[[nodiscard]]` on both `iccBytes()` overloads — the raw-buffer overload returns a caller-owned `new[]` allocation; discarding it is a leak.

WebP scope concern from consistency review: no `SipiIOWebp` handler exists; CLAUDE.md's `TIFF/JPEG/PNG/JP2` claim is complete.

Verification: 8/8 unit tests + 15/15 approval tests still pass twice in a row with `SOURCE_DATE_EPOCH=946684800`.

subotic:
**asan-ubsan failure root-caused and fixed in 73ebc2e.**

The 3-byte leak was a **pre-existing bug** in `Sipi::SipiExif::SipiExif(buf, len)` that my new `PngRoundTrip` approval test exposed for the first time:

```cpp
SipiExif::SipiExif(const unsigned char *exif, unsigned int len)
{
  binaryExif = new unsigned char[len];   // <-- 3 bytes
  memcpy(binaryExif, exif, len);
  binary_size = len;
  try {
    byteorder = Exiv2::ExifParser::decode(exifData, exif, (uint32_t)len);
  } catch (Exiv2::Error &exiverr) {
    throw SipiError(exiverr.what());     // <-- C++: no ~SipiExif on ctor throw → buffer leaks
  }
}
```

The fixture is an SIPI-generated PNG that carries libpng-text-encoded EXIF (3-byte chunk) inherited from the source JPEG. Exiv2 rejects the malformed chunk; SipiError gets thrown; the just-allocated 3-byte `binaryExif` leaks because the constructor never finished, so the destructor doesn't run.

PR 0 didn't trip this because none of its 5 TIFF tests went through `SipiIOPng::read`. My `PngRoundTrip` is the first approval test that does.

The fix holds the buffer in a `std::unique_ptr<unsigned char[]>` through `decode`, then `.release()`s into the raw `binaryExif` member only on success. Stack unwind cleans up the buffer if decode throws.

Verified locally with `build-sanitized`:
- Before: `SUMMARY: AddressSanitizer: 5855 byte(s) leaked in 13 allocation(s)` (incl. the 3-byte one)
- After: `SUMMARY: AddressSanitizer: 5852 byte(s) leaked in 12 allocation(s)` — remaining leaks are macOS-only dyld/ObjC/CoreFoundation init noise that doesn't occur on Linux

CI re-run incoming.

PR #680, opened 2026-06-18
ci: move cross-compile gate to a dedicated amd64 runner (off the darwin gate)

subotic:
[DEV-6563](https://linear.app/dasch/issue/DEV-6563)

## Summary
The darwin-arm64 entry of the `test` job cross-built the Linux image from macOS (Phase 10), adding ~13 min to the most expensive runner. This moves cross-compilation to its own `cross-build` job on a separate **linux-amd64** runner — parallel, and never slows the darwin gate.

## Changes
- **New `cross-build` job** (`ubuntu-24.04`, own concurrency group). From the amd64 exec host it cross-builds the two non-native targets:
  - `linux-arm64`: `//src:image` (`just bazel-cross-build-image arm64`).
  - `darwin-arm64`: `//src/cli:sipi` (`just bazel-build --platforms=//bazel/platforms:darwin_arm64`) — the OCI image is `@platforms//os:linux`-gated, so darwin builds the binary, which still compiles every native dep through the cross toolchain.
- **`//bazel/platforms:darwin_arm64`** platform added (no `darwin_amd64` — the toolchain declares only the `macos-aarch64` target).
- **Removed** the darwin→linux cross-build step from the `test` job.
- Docs (`bazel.md` / `building.md`) + the justfile recipe comment updated to describe the dedicated job.

## Trade-off (worth a look)
This swaps the cross-compile **exec host** from darwin to amd64. We now gate amd64→linux-arm64 and amd64→darwin, and no longer gate darwin→linux. That's the intended move (keep the cross-check off the costly darwin runner); the toolchain is relocatable, so darwin→linux is very likely still fine, but it's no longer CI-verified.

## Verification
- Local: `//src/cli:sipi` builds under `--platforms=darwin_arm64` (✅), linux-arm64 image cross-builds (✅).
- The **amd64→darwin** cross path only runs on a Linux exec host, so its first real validation is this PR's `cross-build` job (same as how the Phase 10 gate self-validated). If `linux→darwin` hits a toolchain gap (macOS SDK / link-wrapper), I'll see it on this PR and adjust.

## Test Plan
- `cross-build` job on this PR exercises both cross targets on a linux-amd64 runner.

## Screenshots
N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,35 +176,6 @@ jobs:
           fi
           nix develop --command bash -c "just bazel-test $REMOTE $BES"
 
-      # darwin→linux cross-compilation gate (DEV-6563 Phase 10). Builds the
-      # Linux OCI image for both target arches FROM the macOS runner, proving
-      # every native cc_library (kakadu/libtiff/exiv2/…) cross-compiles through
-      # the relocatable hermetic clang toolchain with no host-include leaks.
-      # darwin-only: the Linux runners build `//src:image` natively in the
-      # Docker-smoke step, so cross-compilation is the macOS host's job to
-      # prove. Build-only (no `docker load`) — there is no Linux Docker daemon
-      # on the macOS runner.
-      - name: Cross-build Linux image (darwin→linux)
-        if: matrix.platform == 'darwin-arm64'
-        env:
-          GH_TOKEN: ${{ secrets.DASCHBOT_PAT }}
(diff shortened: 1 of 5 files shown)


subotic:
Closing — the dedicated `cross-build` job adds too much CI time (the linux-arm64 image + darwin sipi cross-builds run sequentially on one runner). Parking the cross-compile-in-CI approach to rethink it. Opening a focused PR to also drop the darwin→linux cross-build step (Phase 10) so the darwin gate stays fast; cross-compilation remains available locally via `just bazel-cross-build-image`.

PR #747, opened 2026-07-23
diag: JP2 decode MT disabled at runtime (asan-e2e isolation, do not merge)

subotic:
Controlled experiment for #743. Identical code to that PR but with the Kakadu thread env disabled at runtime (`num_threads = 0`), so decode runs single-threaded. If asan-ubsan-e2e passes here but fails on #743, the runtime multithreading is the cause. Do not merge.


subotic:
Diagnostic complete — closing. This control experiment (MT forced off, otherwise identical to #743) had **asan-ubsan-e2e green**, proving the runtime JP2 decode multithreading is what flipped the gate red on #743. Full root-cause analysis is on #743. Not for merge.

PR #784, opened 2026-08-17
chore(ci): split the review job, gate it on a verified wrong outcome, pin Opus 5 at medium

mpro7:
Rolls the outcome of the 2026-08 review benchmark out to this repo. It ran 157 review runs against 15 real commits with known answers (8 with a planted-and-later-fixed defect, 4 clean, 3 mixed) across dsp-api and dsp-app, and shipped there as dasch-swiss/dsp-api#4252 and dasch-swiss/dsp-app#3336. This repo's workflow is shaped differently, so the port is by substance rather than by copy — what changed and what deliberately did not is below.

## 1. Model: `claude-opus-4-6` → `claude-opus-5`

Model choice dominated prompt choice in the benchmark. Phase 1 scored detection on **two** planted-bug commits across three prompts — **six cells per model**, all run inside the same batches, so nothing else confounds the comparison. **Opus 5 fully caught five of the six and partially caught the sixth, with zero blocking false positives. Opus 4.8 fully caught none** — two partial, four missed. Sonnet 5 fully caught one and produced the only two merge-blocking false positives.

Per-token pricing is **identical** across Opus 5, 4.8, 4.7 and 4.6 — $5.00 in / $25.00 out per MTok — so this is not a cost-for-quality trade on rate. It is worth knowing where a real cost difference does come from: on Opus 5 an omitted `thinking` parameter runs adaptive thinking by default, where Opus 4.8/4.7 run without it, so the same request emits billable thinking tokens. `--effort medium` is the lever for exactly that, and it measured **$3.34 per run against $4.40 at `high`**.

One limit of this evidence, stated because it is what motivated the later prompt work: under those phase-1 prompts **all three models missed a third commit's defect**.

`claude-opus-4-6` is a generation older than the arm that fully caught nothing.

## 2. Effort: pinned to `medium`

Not previously pinned, so runs inherited whatever the action defaults to. Detection was **identical** at `medium` and `high` on all nine commits run at both levels — same catches, same misses, same clean passes. The only behavioural difference is that `high` holds a clean PR shape carrying a verified-real but non-blocking finding: it forms that finding in 3 of 4 runs and blocks in 3 of 3 runs where it formed it. `medium` never produced a wrong merge call in ten runs and costs ~16% less.

## 3. The review decision now keys on evidence, not on a category

This is the substantive change, and it is the same defect the benchmark found in dsp-api — arrived at independently here.

The gate was a taxonomy: 🔴 Request changes for *"security issues, memory safety violations, heap-buffer-overflow risks, type mismatches at C library boundaries"*. That has the failure mode measured repeatedly in dsp-api's Critical-only rule — the gate rides on a label rather than on what the review actually established. Two consequences follow directly:

- a **speculative** memory-safety worry outranks a **verified** wrong-output bug that happens to fall outside the listed categories;
- a reviewer that grades a defect *accurately* can be obliged to let it through, while one that over-grades blocks it. In dsp-api that was observed directly: the one configuration that graded a real defect exactly as the answer key does was the one forced to clear, and the only way to block was to inflate the severity.

Findings are now **Blocking** or **Non-blocking**, independent of how bad the outcome is and of which REVIEW.md section it came from. Blocking means the review *established* — from the code or an authoritative external source — that a reachable input or state produces a wrong outcome: memory unsafety, an error or crash on a normal path, incorrect bytes or dimensions written or returned, access incorrectly granted or denied, or a deterministic CI failure. Impact (High/Medium/Low) survives for triage ordering only and never decides the gate. 🔴/🟡/🟢 then follow mechanically.

Measured effect of that rewrite in dsp-api and dsp-app: **10 of 10** correct merge calls at `medium` across nine commits, with **zero** false blocks on four clean commits and **zero** false positives on deliberately-documented-as-intended code — against a gate failure in every prior configuration.

One line closes the failure the benchmark caught most often: *if you find yourself arguing that a Non-blocking issue should hold the PR, it is Blocking; re-mark it.* Reviews really did write "ready to merge" and then argue for a fix in the same paragraph.

## 4. Two prompt clauses added, adapted to this stack

**Trigger and wrong outcome, or drop it.** Every reported issue must name the concrete input or state that triggers it and the wrong output, crash or regression that results. Anything justified only as *"more robust"* or *"best practice"*, or already explained as deliberate in a code comment, is dropped. An unproven-but-real-looking concern is reported as `Suspected` rather than dropped or inflated — always Non-blocking.

**Do not assert external behaviour from memory.** When the diff constructs or parses a value another system interprets — an IIIF size, region or rotation parameter, a URL, a path template, an image-format header, a media-type string — check the constructed form against that system's own specification before reporting *or clearing* it. Well-formed is not the same as accepted: a form can be valid syntax, listed as supported, and still rejected at runtime by a precondition attached to it. Confirm against `MODULE.bazel` or the specification itself; if the check cannot be completed, say so and grade `Suspected`.

That second clause is the highest-leverage sentence in the whole benchmark: one bug class went from **16 straight misses to 2 of 2** when it was added, and the bug in question was a request this server rejects with a 400. It applies more directly here than anywhere else.

Repo-specific emphases are kept as they were: REVIEW.md as the checklist (now with its "Skip" section honoured explicitly), the FFI and multi-buffer-dimension sections called out, sanitizer CI status folded into the review, and inline comments through a single review. General `@claude` mentions are still served, now by the separate `claude-general` job rather than by a dispatch branch inside this prompt.

## 5. The review is split out of the shared job, and its tool set is aligned

The review and the general assistant shared a single `claude` job, so every tool decision was a trade-off between two different jobs: leave the tools open and the review cannot be pinned to what was benchmarked; close them and `@claude <prompt>` loses the ability to edit code. This PR splits them, which removes the trade-off and brings this repo in line with dsp-api, dsp-app and dsp-repository.

**`claude-code-review`** (new) carries the benchmarked configuration:

```
--model claude-opus-5 --effort medium --setting-sources user
--allowed-tools "Read,Grep,Glob,WebSearch,WebFetch,Bash(gh …)"
```

- **`Read` has to be granted explicitly.** `--allowed-tools` is the complete resolved set — the action adds nothing to it, verified by reading a real CI run's `SDK options:` echo. Without it, the prompt's instruction to read the changed files in full cannot be honoured.
- **`WebFetch`/`WebSearch` are what make the external-verification clause executable at all.** This is the gap the first commit left open: it added a clause requiring external verification without confirming the reviewer could fetch anything. In this repo that mattered more than elsewhere — the clause exists to catch preconditions like the 400 responses this server itself returns.
- **Three tools beyond the dsp-api list**, each required by this repo's prompt rather than by preference: `Bash(gh api:*)` and `Bash(gh pr review:*)` to submit one inline review, and `Bash(gh run …)` to read sanitizer CI status.
- **`--setting-sources user`** stops the job inheriting this repo's `.claude/settings.json` — see the note below, which this now resolves.
- **Permissions drop to `contents: read`** plus `pull-requests: write` and `actions: read`. The shared job needed `contents: write` for the assistant's sake; a review does not.

**`claude-general`** keeps its write permissions, its full-history checkout and an unrestricted tool set, and is gated as exactly the negation of the review condition, so the two jobs are mutually exclusive by construction. Only its model moves to Opus 5 — none of the benchmark's findings concern that arm.

## What this PR deliberately does not change

**The general assistant's behaviour.** Its tools stay unrestricted, its permissions stay as they were, and it keeps `fetch-depth: 0`. The only change is the model.

**The DEV-6884 author gate.** Preserved on both jobs, with the general arm keeping the per-trigger-arm form so an `issues` payload cannot slip through on `issue.author_association`.

## Two things worth a look, separately from this PR

1. **CI inherits this repo's `.claude/settings.json`.** The action loads `settingSources: user,project,local`, and CI restores that file from the default branch before running — verified by reading a real run's `SDK options:` echo in dsp-api. This repo's file grants `docker run:*`, `gcloud storage *`, `gcloud projects *`, `gcc`, `g++`, `find` and `grep` to any session that loads it, including this review job, which also holds `contents: write`. Those grants are a local-developer convenience list, not something written for CI. **Resolved for the review job in this PR** via `--setting-sources user`. It is still inherited by `claude-general`, deliberately: that arm is interactive and may want those grants. Worth deciding separately whether it should be.

2. **`dasch-swiss/ark-resolver` has a Claude review workflow that the DEV-6884 sweep missed.** Its file is `.github/workflows/claude-review.yml`, and it contains **no** `author_association` check — the exact gap DEV-6884 closed in the other five repos, still open on the same API key. Found by probing all 155 non-archived org repos for a `claude*` workflow; that probe returns six repos, where DEV-6884's attachments list five.


.github/workflows/claude.yml
```diff
@@ -11,62 +11,198 @@ on:
...
+      actions: read
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - uses: anthropics/claude-code-action@v1
+        with:
+          anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          claude_args: "--model claude-opus-5"
```
subotic, line 208: I would prefer opus 4.8. I don't use Opus 5 as it is to expensive for what it is.

subotic (approved):

.github/workflows/claude.yml
mpro7: @subotic I can change it to Opus 4.8 if you really want to, but model choice dominated prompt choice in the benchmark. Phase 1 scored detection on two planted-bug commits across three prompts. Six cells per model, all run inside the same batches, so nothing else confounds the comparison. Opus 5 fully caught five of the six and partially caught the sixth, with zero blocking false positives. Opus 4.8 fully caught none., two partial and four missed. Sonnet 5 fully caught one and produced the only two merge-blocking false positives.

Per-token pricing is identical across Opus 5, 4.8, 4.7 and 4.6 — $5.00 in / $25.00 out per MTok — so this is not a cost-for-quality trade on rate. It is worth knowing where a real cost difference does come from: on Opus 5 an omitted thinking parameter runs adaptive thinking by default, where Opus 4.8/4.7 run without it, so the same request emits billable thinking tokens. `--effort` medium is the lever for exactly that, and it measured $3.34 per run against $4.40 at high. This are still benchmark numbers only tested on limited number of factors, but IMO reliable. You can [read more here](https://claude.ai/code/artifact/8a4a4e91-f1d2-4b69-aeee-5914728768f7?org=a6b9bab6-28b0-4f89-8d01-c811b292f94c) about the methodology, test data and results. 

Let me know if I should change the model here.

subotic: thanks, you can leave it at Opus 5

subotic (approved):

